PR #27, opened 2024-01-30
chore: fix pip install in contributing md

lsattolo:
`pip install --editable .[tests]` gives me: `no matches found: .[tests]`


HugoDelatte (approved):
Thanks for PR! 
Indeed, when using Unix Shell you have to escape the square brackets as opposed to Windows.  So by using string, we ensure that it is working for everyone.

PR #57, opened 2024-06-03
[BUG] feat(prices_to_returns): Add index intersect param

### LINKED ISSUE #55: prices_to_returns cuts off dataframe to last asset launch [BUG]
**Describe the bug**

`prices_to_returns` promises to take a dataframe of prices and convert them to returns. In the event I have a dataframe with returns with different starting times (such as `VTI` and `BITI`) this function will cut rows to the start of `BITI`.

**To Reproduce**

```python
import yfinance as yf
import pandas as pd

STOCKS = ["VTI", "BITI"]
return_dfs = []
for ticker in STOCKS:
  ticker_yahoo = yf.Ticker(ticker)
  df = ticker_yahoo.history(period="1d", start="2010-01-01")
  df.index = df.index.date
  return_dfs.append(df[["Close"]].rename(columns={"Close": ticker}))
return_df = pd.concat(return_dfs, axis=1).sort_index().fillna(0.0)

prices_to_returns(return_df)
```

**Expected behavior**

Returns for `BITI` before `2022-06-21` are 0.0. Instead it cuts off the dataframe.

**Additional context**

This occurs due to the [dropna](https://github.com/skfolio/skfolio/blob/67a2a18288fcecd5f911d67402e6839b0332fa5a/src/skfolio/preprocessing/_returns.py#L107). I believe this should use `fillna` with 0.0 after the `pct_change` call to mark the previous periods are 0.0 for returns. I would be more than happy to make this change just want to be sure this is considered the right move by the maintainers.

**Versions**

`0.2.1`

8W9aG:
* Add a parameter allow index intersection to be chosen by the caller.
* This allows the dropna call to either use all or any.
* Functionally this determines whether the start of the index is at the beginning of the shortest price index (any) or the beginning of the longest price index (all)

#### Reference Issues/PRs

Fixes #55 

Allows the user to choose whether to dropna's using `any` or `all` to preserve the full time index.

No.

Whether this fix conforms to the discussion in the issue.

I extended the `test_returns` test to check for this.

#### Any other comments?

The original behaviour should be unchanged.

##### For all contributions
- [x ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.
- [ ] Optionally, I've added myself and possibly others to the [CODEOWNERS](https://github.com/skfolio/skfolio/blob/main/CODEOWNERS) file - do this if you want to become the owner or maintainer of an estimator you added. -- File does not exist

##### For new estimators

-- No new estimators added.

### DIFF
--- a/src/skfolio/preprocessing/_returns.py
+++ b/src/skfolio/preprocessing/_returns.py
@@ -14,7 +14,7 @@ def prices_to_returns(
     log_returns: bool = False,
     nan_threshold: float = 1,
     join: str = "outer",
-    index_intersect: bool = True,
+    drop_inceptions_nan: bool = True,
 ) -> pd.DataFrame | tuple[pd.DataFrame, pd.DataFrame]:
     r"""Transforms a DataFrame of prices to linear or logarithmic returns.
 
@@ -108,7 +108,7 @@ def prices_to_returns(
     df.dropna(axis=1, how="all", inplace=True)
 
     # returns
-    all_returns = df.pct_change().dropna(how="any" if index_intersect else "all")
+    all_returns = df.pct_change().dropna(how="any" if drop_inceptions_nan else "all")
     if log_returns:
         all_returns = np.log1p(all_returns)
 
--- a/src/skfolio/preprocessing/_returns.py
+++ b/src/skfolio/preprocessing/_returns.py
@@ -61,8 +61,8 @@ def prices_to_returns(
         Drop observations (rows) that have a percentage of missing assets prices above
         this threshold. The default (`1.0`) is to keep all the observations.
 
-    index_intersect : bool, default=True
-        Use the intersection of the index to determine when the dataframe index begins.
+    drop_inceptions_nan : bool, default=True
+        If this is set to True,  observations at the beginning are dropped if any of the asset values are missing, otherwise we keep the NaNs. This is useful when you work with a large universe of assets with different inception dates coupled with a pre-selection Transformer.
 
     Returns
     -------
--- a/src/skfolio/preprocessing/_returns.py
+++ b/src/skfolio/preprocessing/_returns.py
@@ -14,6 +14,7 @@ def prices_to_returns(
     log_returns: bool = False,
     nan_threshold: float = 1,
     join: str = "outer",
+    index_intersect: bool = True,
 ) -> pd.DataFrame | tuple[pd.DataFrame, pd.DataFrame]:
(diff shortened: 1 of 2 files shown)


src/skfolio/preprocessing/_returns.py
```diff
@@ -14,6 +14,7 @@ def prices_to_returns(
     log_returns: bool = False,
     nan_threshold: float = 1,
     join: str = "outer",
+    index_intersect: bool = True,
```
HugoDelatte, line 17: I was thinking about `drop_inceptions_nan` to be very explicit

tests/test_preprocessing/test_returns.py
```diff
@@ -37,3 +37,8 @@ def test_returns(prices, factor_prices):
     assert np.all(X.columns == prices.columns)
     assert np.all(y.columns == factor_prices.columns)
     assert np.all(X.index == y.index)
+
+    # Test index_intersect by making the first column mostly 0's
+    prices[prices.columns.values[0]].iloc[:-10] = 0.0
```
HugoDelatte, line 42: to avoid the pandas warning on new version, let's replace by:
```suggestion
    prices.loc[:prices.index[-10], "AAPL"] = 0.0
```

src/skfolio/preprocessing/_returns.py
```diff
@@ -60,6 +61,9 @@ def prices_to_returns(
         Drop observations (rows) that have a percentage of missing assets prices above
         this threshold. The default (`1.0`) is to keep all the observations.
 
+    index_intersect : bool, default=True
+        Use the intersection of the index to determine when the dataframe index begins.
+
```
HugoDelatte, line 66:
```suggestion
    drop_inceptions_nan : bool, default=True
        If this is set to True,  observations at the beginning are dropped if any of the asset values are missing, otherwise we keep the NaNs. This is useful when you work with a large universe of assets with different inception dates coupled with a pre-selection Transformer.

```

HugoDelatte (changes requested):
Thank you for your PR!
That's indeed to correct solution. I've just added some comments on the naming and tests.
After that, it can be merged

src/skfolio/preprocessing/_returns.py
8W9aG: Done

HugoDelatte (approved):

PR #61, opened 2024-06-09
feat: add yields_to_returns function to calculate bond market invariants

Conway-Yu:
#### Reference Issues/PRs

Added market invariant estimation for bonds using yields and maturities.

As using prices does not capture the total return of a bond and coupon data is often hard to obtain, we use the following approximation method from [https://www.mdpi.com/2306-5729/4/3/91](url) to estimate total return via yield and maturity.
No. 

Checking if my implementation can be improved on and if it follows generally the skfolio API logic.

No, but one can easily use the dataset in [https://datarepository.eur.nl/articles/dataset/Data_Treasury_Bond_Return_Data_Starting_in_1962/8152748](url) and compare the results of the function to the "official" total returns.

#### Any other comments?

The approximation method also works with corporate bond data and one can reasonably good results.

##### For all contributions
- [ ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

##### For new estimators

### DIFF
--- a/src/skfolio/preprocessing/_returns.py
+++ b/src/skfolio/preprocessing/_returns.py
@@ -124,3 +124,200 @@ def prices_to_returns(
     returns = all_returns[[x for x in X.columns if x in df.columns]]
     factor_returns = all_returns[[x for x in y.columns if x in df.columns]]
     return returns, factor_returns
+
+
+def yields_to_returns(
+    X: pd.DataFrame,
+    X_maturity: pd.DataFrame,
+    y: pd.DataFrame | None = None,
+    y_maturity: pd.DataFrame | None = None,
+    log_returns: bool = False,
+    nan_threshold: float = 1,
+    join: Literal["left", "right", "inner", "outer", "cross"] = "outer",
+    drop_inceptions_nan: bool = True,
+) -> pd.DataFrame | tuple[pd.DataFrame, pd.DataFrame]:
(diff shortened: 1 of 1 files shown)


HugoDelatte:
Thank you for your PR! I'll review it this week and let you know if modifications are needed

Conway-Yu:
Hello,
To test this approximation I used the proposed scatter and histogram tests by [Meucci](https://www.arpm.co/symmys-articles/4%20_%20Attilio%20Meucci%20-%20Risk%20and%20Asset%20Allocation%20-%20Quest%20for%20Invariance.pdf), and the plots were satisfactory.

However, I believe my approach is note quite correct in every scenario and I also found another approach using fictive zero coupon curves in [here](https://www.pm-research.com/content/iijfixinc/31/4/50). Therefore, I'll abolish the pull request for now to rework my proposal.

HugoDelatte:
Thank you for the update, I was about to review it (now that the Metadata routing PR has been finished).
Having specific tools to work with bonds inside `skfolio` is on our roadmap.
`skfolio` is a toolkit so it's fine to have multiple methods to do the same thing, but if one is clearly superior we should indeed implement it.
Feel free to submit the second method if you wish, and thanks again for your first PR.

PR #97, opened 2024-11-16
[MNT] Pin `kaleido==0.2.1` version to fix `sphinx-build`

peter-gy:
#### What does this implement/fix? Explain your changes.

The documentation site is built using Sphinx, in turn making use of `plotly_sg_scraper` under the hood to render plot images, which itself in turn depends on [kaleido](https://pypi.org/project/kaleido/). `kaleido 0.4.1` got released recently (on Nov 15, 2024) introducing some changes under the hood causing the `sphinx-build . _build` command to run indefinitely.

This PR pins the version of `kaleido` to `0.2.1`, the version that has been resolved by the dependency graph prior to Nov 15, 2024.

#### Does your contribution introduce a new dependency? If yes, which one?

N/A
#### What should a reviewer concentrate their feedback on?

The `Deploy Documentation` GitHub workflow should keep working as expected.

#### Did you add any tests for the change?

N/A

##### For all contributions

- [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.
- [ ] Optionally, I've added myself and possibly others to the [CODEOWNERS](https://github.com/skfolio/skfolio/blob/main/CODEOWNERS) file - do this if you want to become the owner or maintainer of an estimator you added.

##### For new estimators

N/A

PR #98, opened 2024-11-16
[ENH] Integrate JupyterLite into documentation site

peter-gy:
Builds on top of #97

This PR integrates [JupyterLite](https://jupyterlite.readthedocs.io/) into the documentation site, allowing visitors of the docs to try out examples in their browsers **without Binder**, thanks to a WebAssembly-powered Python kernel running in the browser ([Pyodide Kernel](https://pyodide.org/en/stable/)).

Also modernizes the `docs.yml` workflow by using actions maintained by GitHub.

You can take a look at the product of this PR here: https://peter-gy.github.io/skfolio.
Example with "launch lite" button: https://peter-gy.github.io/skfolio/auto_examples/1_mean_risk/plot_1_maximum_sharpe_ratio.html#conclusion

https://github.com/user-attachments/assets/5a95e5e9-8e65-4309-98c1-73f5bfb93b6b

This work only introduces dependencies exclusively to the `docs` optional group:

- [jupyterlite-sphinx](https://pypi.org/project/jupyterlite-sphinx/): wraps the JupyterLite CLI into a Sphinx Extension
- [jupyterlite-pyodide-kernel](https://pypi.org/project/jupyterlite-pyodide-kernel/): provides us with WebAssembly-based Python kernel, exposing dependencies such as `scikit-learn`, `pandas`, `cvxpy`, etc. compiled into WASM via CPython & Emscripten
- [nbformat](https://pypi.org/project/nbformat/): to programmatically mutate Jupyter notebook files

The monkey patching of Jupyter Notebooks in `docs/conf.py#patch_jupyterlite_notebooks`.

For the record, the nicest solution would be to maintain a custom Pyodide distribution with `skfolio` and `plotly` being members of it by default. However, setting up and maintaining the associated CI/CD pipeline would be a real headache. (I roughly outlined what steps would be necessary for that here: https://github.com/pyodide/pyodide/pull/5183#issuecomment-2471628369)

N/A

##### For all contributions


HugoDelatte:
Really really cool stuff, well done!
Integrating JupyterLite is something I wanted to do for a long time so thanks for that PR.
I'll take the time to review it properly tomorrow.

HugoDelatte:
The PR looks good. 
I'm gonna release a new feature including two examples that use a mix-integer programing solver (SCIP).
SCIP is an optional dependency of cvxpy, that can be installed using  `pip install cvxpy[SCIP]` or `pip install PySCIPOpt`.
I'll add this SCIP package only in the docs for the example to build.
However, looks like PySCIPOpt doesn't have a pure Python wheel to work in JupyterLite.
My question is, what would be the best strategy to handle that? 
I guess one solution would be to remove the JupyterLite button for specific example?

peter-gy:
Whenever we need a dependency that has no pure Python wheel, we need to check whether Pyodide maintains a build recipe for it under https://github.com/pyodide/pyodide/tree/main/packages.

In this case, `PySCIPOpt` is not present, so indeed, features requiring that as a dependency will not work in JupyterLite.
However, instead of disabling the JupyterLite button, a better strategy would be to open a PR at Pyodide to add support to `PySCIPOpt`, the maintainers are really receptive to admitting new packages.

HugoDelatte:
The doc [deployment failed](https://github.com/skfolio/skfolio/actions/runs/11880630938/job/33104009314)
I'll investigate further

peter-gy:
> The doc [deployment failed](https://github.com/skfolio/skfolio/actions/runs/11880630938/job/33104009314) I'll investigate further

I believe #99 should fix the issue.

HugoDelatte:
Just noticed that some example that require an external data download fail like for [this one](https://skfolio.org/auto_examples/1_mean_risk/plot_8_regularization.html#sphx-glr-auto-examples-1-mean-risk-plot-8-regularization-py).

The function `load_ftse100_dataset` download data from the skfolio-dataset repo. We did that on some datasets to reduce the main repo and package size.

Any clue on how to fix this?

peter-gy:
> Just noticed that some example that require an external data download fail like for [this one](https://skfolio.org/auto_examples/1_mean_risk/plot_8_regularization.html#sphx-glr-auto-examples-1-mean-risk-plot-8-regularization-py).
> 
> The function `load_ftse100_dataset` download data from the skfolio-dataset repo. We did that on some datasets to reduce the main repo and package size.
> 
> Any clue on how to fix this?

Sure.

This is caused by the fact that when using JupyterLite, everything happens on the client side, within our browser, including the HTTP request we make in [download_dataset](https://github.com/skfolio/skfolio/blob/dc5671f3e9da3595136d5e3f88028fa8e19b0d3a/src/skfolio/datasets/_base.py#L143-L158). Our browser requests `https://github.com/skfolio/skfolio-datasets/raw/main/*`, but GitHub servers will block it because of CORS issues. (cc. https://github.com/pyodide/pyodide/issues/4060)

I see two solutions to get around that.

(1) The quick and pragmatic solution is to use a CORS proxy such as https://corsproxy.io/. We can modify `download_dataset` to check `if sys.platform == 'emscripten'` indicating at runtime that skfolio is being called from a browser. If that's the case, we prepend `https://corsproxy.io/?` to the dataset URL. Super simple solution at the app level, no need to mess with setting up infrastructure.

(2) We start hosting `skfolio-dataset` contents under a blob storage (AWS S3, Google Cloud Storage, Azure Bucket, etc.) where we have control over CORS policy and modify `download_dataset` to point to those URLs. This is a more heavy-weight solution, but has the benefit that the core source code would not need to include `sys.platform` check.

Due to its simplicity, I would suggest to go with option (1) for now.

HugoDelatte:
Was still failing but found the solution by calling (https://github.com/jupyterlite/pyodide-kernel/issues/109):
 ```python
import pyodide_http
pyodide_http.patch_all()
 ```
 
 I'll add that in the JupyterLite Initialization section.

PR #103, opened 2024-11-30
feat: _max_empirical_covariance

matteoettam09:
##### For all contributions
- [ ] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

##### For new estimators

### DIFF
--- a/src/skfolio/moments/covariance/_empirical_covariance.py
+++ b/src/skfolio/moments/covariance/_empirical_covariance.py
@@ -9,6 +9,7 @@
 
 import numpy as np
 import numpy.typing as npt
+from typing import Literal
 
 from skfolio.moments.covariance._base import BaseCovariance
 
@@ -62,7 +63,7 @@ class EmpiricalCovariance(BaseCovariance):
 
     def __init__(
         self,
-        window_size: int | None = None,
+        window_size: int | None | Literal["max"] = None,
         ddof: int = 1,
         nearest: bool = True,
         higham: bool = False,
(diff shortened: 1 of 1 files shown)


HugoDelatte:
Thank you for the PR!

You’re addressing an important point with deep ramifications. A full resolution will take some time, but it will unlock some of the requested features.

We indeed need methods to estimate covariance when data are missing, using overlapping (pairwise-complete) observations. _(Additionally, we may want to apply the Expectation–Maximization algorithm, since in large cross-sectional sets, the naive approach seems to lead to less coherent covariance matrices.)_

More generally, we want to apply these methods for all relevant moments (e.g., mean, covariance, etc.) and potentially use the newly estimated distributions to generate returns.

Some users have already expressed the need for portfolio construction under generative returns, especially when the dataset is small (this is particularly relevant for CVaR).

The API was designed with this possibility in mind. The current workflow is as follows:

The returns are consumed by a PriorEstimator.
This estimator estimates the distribution and create a PriorModel that contains the estimated moments (mean and covariance), the Cholesky decomposition, and the estimated returns.
This PriorModel is then consumed by the Optimization/Allocation estimators.

In most cases, the estimated returns are the same as the input (except for factor models), but nothing prevents us from returning generative returns at this step if we want.

It just needs some additional methods/NaNs handling.
I’ll start working on this. Once it’s done, we’ll be able to integrate your covariance estimator.

Issue #131, opened 2025-03-23
[ENH] Pre-selection transformer to drop near zero variance assets


HugoDelatte:
Add a pre-selection transformer to drop near zero variance assets as discussed in https://github.com/skfolio/skfolio/discussions/130

PR #132, opened 2025-03-30
[ENH] add DropZeroVariance

Vincent-Maladiere:
This is my first PR with skfolio! :)

#### Reference Issues/PRs

Fixes https://github.com/skfolio/skfolio/issues/131

A simple `DropZeroVariance` transformer to drop assets with near-zero variance.

No

Yes

##### For all contributions
	-> this file doesn't exist anymore, should we remove this entry in the PR template?

##### For new estimators
- [x] I've added the estimator to the API reference in `docs/api.rst`.
- [x] I've added one or more illustrative usage examples to the docstring and the `examples` section.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
     hooks:
       - id: ruff
         name: ruff-lint
-        args: [check, --fix, --exit-non-zero-on-fix]
+        args: [--fix, --exit-non-zero-on-fix]
       - id: ruff
         name: ruff-format
         args: [format, --diff]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,6 @@
 
 * fix(population): surface plot compatible with plotly v6.0 ([`c053625`](https://github.com/skfolio/skfolio/commit/c053625d641eda1deee480aafd8d10cd0e6b3c56))
 
-### Feature
-
-* feat: DropZeroVariance estimator
-
 ## v0.8.0 (2025-03-21)
 
 ### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 * fix(population): surface plot compatible with plotly v6.0 ([`c053625`](https://github.com/skfolio/skfolio/commit/c053625d641eda1deee480aafd8d10cd0e6b3c56))
 
+### Feature
+
+* feat: DropZeroVariance estimator
+
 ## v0.8.0 (2025-03-21)
 
 ### Breaking
--- a/src/skfolio/pre_selection/_drop_zero_variance.py
+++ b/src/skfolio/pre_selection/_drop_zero_variance.py
@@ -1,0 +1,68 @@
+"""Pre-selection DropZeroVariance module."""
+
+import numpy as np
+import numpy.typing as npt
+import sklearn.base as skb
+import sklearn.feature_selection as skf
+import sklearn.utils.validation as skv
+
+
+class DropZeroVariance(skf.SelectorMixin, skb.BaseEstimator):
+    """Transformer for dropping assets with near-zero variance.
+
+    On short windows, some assets can experience a near-zero variance, making
+    the covariance matrix improper for optimization. This simple transformer drops
+    assets whose variance is below some threshold.
+
+    Parameters
+    ----------
+    threshold : float, default=1e-6
+        Correlation threshold. The default value is `1e-6`.
+
+    Attributes
+    ----------
+    to_keep_ : ndarray of shape (n_assets, )
+        Boolean array indicating which assets are remaining.
+
+    n_features_in_ : int
+        Number of assets seen during `fit`.
+
+    feature_names_in_ : ndarray of shape (`n_features_in_`,)
+        Names of assets seen during `fit`. Defined only when `X`
+        has assets names that are all strings.
+    """
+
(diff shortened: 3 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ repos:
     hooks:
       - id: ruff
         name: ruff-lint
-        args: [check, --fix, --exit-non-zero-on-fix]
```
Vincent-Maladiere, line 16: I fought with the linting for a while, and removing this argument solved my issue. WDYT?

HugoDelatte: This lines from the pre-commit hook run a ruff check and fix linting issues then raise an error if it did encountered some issues so that the user is made aware.
If you did run `ruff check --fix` and `ruff format` before commiting, you shouldn't have any issues.
The only rare case where you can have an issue is when your local version of ruff is not up-to-date and the latest version in the pipeline has some new rule --> maybe try to update ruff and see if it resolve the issue?

CHANGELOG.md
HugoDelatte, line 1: You don't have to amend this file, it's done automatically in the github workflow.

src/skfolio/pre_selection/_drop_zero_variance.py
```diff
@@ -0,0 +1,68 @@
...
+import numpy.typing as npt
+import sklearn.base as skb
+import sklearn.feature_selection as skf
+import sklearn.utils.validation as skv
+
+
+class DropZeroVariance(skf.SelectorMixin, skb.BaseEstimator):
+    """Transformer for dropping assets with near-zero variance.
+
+    On short windows, some assets can experience a near-zero variance, making
```
HugoDelatte, line 13: This description is spot on! 👌

src/skfolio/pre_selection/_drop_zero_variance.py
```diff
@@ -0,0 +1,68 @@
...
+    the covariance matrix improper for optimization. This simple transformer drops
+    assets whose variance is below some threshold.
+
+    Parameters
+    ----------
+    threshold : float, default=1e-6
```
HugoDelatte, line 19: I would put the default to 1e-8
The rational is that a variance of 1e-6 means a std of 1e-3 so 0.10% per day (for daily X) which is around 1.6% per annum. Some very low volatility  assets can actually be considered valid and have such low volatility.

tests/test_pipeline/test_pipeline.py
HugoDelatte, line 1: Indeed, much cleaner!

HugoDelatte (changes requested):
Thank you for the PR, it's top work!

src/skfolio/pre_selection/_drop_zero_variance.py
HugoDelatte, line 1: I usually put the following on top of the file, WDYT?

```python
# Copyright (c) 2025
# Author: <your name>
# SPDX-License-Identifier: BSD-3-Clause
```

src/skfolio/pre_selection/_drop_zero_variance.py
Vincent-Maladiere: That's great insights!

src/skfolio/pre_selection/_drop_zero_variance.py
Vincent-Maladiere: I don't have a strong opinion, but I think consistency is important, so I'll add it indeed :)

.pre-commit-config.yaml
Vincent-Maladiere: Ok, I think I understand what's going on. This pre-commit config is slightly incorrect because:
- `id: ruff` calls the [`ruff` hook](https://github.com/astral-sh/ruff-pre-commit/blob/main/.pre-commit-hooks.yaml#L1) whose entry command is `ruff check`, not just `ruff`.
- `args: [check, --fix, --exit-non-zero-on-fix]` passes arguments to `ruff check`, so here it's equivalent to running `ruff check check --fix --exit-non-zero-on-fix`, which yields
  ```
   check:1:1: E902 No such file or directory (os error 2)
  ```
  due to the extra `check`
- The same goes for formatting, `id: ruff`, with `args: [format, --diff]` will result in `ruff check format --diff`. What you want here is using the [`ruff-format` hook](https://github.com/astral-sh/ruff-pre-commit/blob/main/.pre-commit-hooks.yaml#L12), using `id: ruff-format` with `args: [--diff]`, which is equivalent to running `ruff format --diff`
- You haven't run into this issue because you probably haven't installed this pre-commit config, and preferred running these ruff commands yourself? I'm just guessing here :)

All in all, this gives the following ruff config in your .pre-commit-config.yaml
```yaml
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.8.4
    hooks:
      - id: ruff
        name: ruff-lint
        args: [--fix, --exit-non-zero-on-fix]
      - id: ruff-format
        name: ruff-format
        args: [--diff]

```
Let me know what you think.

Vincent-Maladiere: (I can create a separate PR for this if that makes sense)

HugoDelatte: You are correct according to https://github.com/astral-sh/ruff-pre-commit
No idea why it didn't fail on my side and also the others as I get a notification that it's running...
Let's fix it in this PR as it's benign

HugoDelatte (approved):
All good, thx for this PR!

PR #157, opened 2025-07-02
[ENH] Added dockerfile and instructions to run jupyterlab inside docker

CarloNicolini:
I’ve added a Dockerfile to allow running skfolio-jupyterlab from within Docker. This is intended for users who encounter issues installing it directly on their systems.
The Docker image is based on Python 3.12, and skfolio is installed using Astral’s uv package installer.

Also added a small section in README.rst to help users mount a local folder from the docker jupyterlab so that users can run the library without installing it locally in case they have compatibility issues.

Nothing changed in pyproject.toml

Reviewing the Dockerfile to check it conforms to docker best practices.

No tests are required for this addition.

#### Any other comments?
This should make it easier for users who face environment or dependency issues to get started quickly with skfolio in JupyterLab.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,21 +82,8 @@ Ready to contribute? Here's how to set yourself up for local development.
    $ uv run pytest
    ```
 
-6. Then run linting and formatting checks with:
+6. Then run linting and formatting with:
 
-   Using **pip**:
-   ```shell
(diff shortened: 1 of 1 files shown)


HugoDelatte (approved):
I've corrected the tests and changed a bit the doc.
Thanks for the PR!

PR #184, opened 2025-10-09
[ENH] automated code generation with context7 MCP server and rules

moyuanchen:

Added mcp server configuration file and context7 rule file for using cursor agent to automatically generate a portfolio for the user when prompted. 

Some examples of usage in agent mode and plan mode is included.

See examples/mcp/

##### For all contributions

##### For new estimators

### DIFF
--- a/context7.json
+++ b/context7.json
@@ -1,7 +1,7 @@
 {
     "$schema": "https://context7.com/schema/context7.json",
     "projectTitle": "skfolio",
-    "description": "A Python library for portfolio optimization, compatible with scikit-learn.",
+    "description": "skfolio is a Python library for portfolio optimization and risk management built on top of scikit-learn.",
 
     "folders": [
       "docs", 
@@ -34,7 +34,6 @@
     "rules": [
       "Input data (X) should be asset price returns, not prices.",
       "Follows scikit-learn API: use .fit(X, y) to train and .predict(X) to get weights.",
-      "Ensure data is pre-processed and does not contain NaNs before fitting.",
-      "The final portfolio weights are in the .weights_ attribute after fitting."
+      "Ensure data is pre-processed and does not contain NaNs before fitting."
     ]
   }
--- a/context7.json
+++ b/context7.json
@@ -4,12 +4,12 @@
     "description": "skfolio is a Python library for portfolio optimization and risk management built on top of scikit-learn.",
 
     "folders": [
-      "docs", 
-      "src/skfolio"
+      "docs"
     ],
   
     "excludeFolders": [
       "tests",
+      "src",
       "**/tests",
       "*.test",
       "__pycache__",
--- a/context7.json
+++ b/context7.json
@@ -1,0 +1,40 @@
+{
+    "$schema": "https://context7.com/schema/context7.json",
+    "projectTitle": "skfolio",
+    "description": "A Python library for portfolio optimization, compatible with scikit-learn.",
+
+    "folders": [
+      "docs", 
+      "src/skfolio"
+    ],
+  
+    "excludeFolders": [
+      "tests",
+      "**/tests",
+      "*.test",
+      "__pycache__",
+      "*.egg-info",
+      "build",
+      "dist",
+      ".venv",
+      "venv",
+      "examples",
+      "notebooks"
+    ],
+  
+    "excludeFiles": [
+       "CONTRIBUTING.md",
+       "Dockerfile",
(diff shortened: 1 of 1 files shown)


HugoDelatte:
Thanks for the PR.
Correct me if I'm wrong, but .cursor/mcp.json file is for direct local dev of skfolio so I don't think we should ship it in the library repo.

What we probably want is better context for all users of skfolio, for example using this: https://github.com/upstash/context7/blob/master/docs/adding-projects.md#advanced-configuration-with-context7json

WDYT?

context7.json
```diff
@@ -0,0 +1,39 @@
+{
+    "$schema": "https://context7.com/schema/context7.json",
+    "projectTitle": "skfolio",
+    "description": "skfolio is a Python library for portfolio optimization and risk management built on top of scikit-learn.",
+
+    "folders": [
+      "docs", 
+      "src/skfolio"
+    ],
```
HugoDelatte: Add also the examples:

```

  "folders": [
    "src/skfolio",
    "examples",
    "docs"
  ],
  ```

File: context7.json [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+       "pyproject.toml",
+       "SECURITY.md"
+    ],
+
+    "rules": [
+      "Input data (X) should be asset price returns, not prices.",
+      "Follows scikit-learn API: use .fit(X, y) to train and .predict(X) to get weights.",
+      "Ensure data is pre-processed and does not contain NaNs before fitting."
+    ]
+  }
```
HugoDelatte: I would live the rules empty and let the LLM figure it out based on context

HugoDelatte:
It seems context7 doesn't read python files, only .rst, .md, or .txt
So we are missing out on the examples and API as they are generated at build time and into HTML (which isn't optimal either anyway).

Will need to dig deeper to find a clean solution (like creating a dedicated contect7 branch and building API+examples into .txt).

Or maybe it's just easier to make our own context/MCP.

WDYT?

CarloNicolini:
> It seems context7 doesn't read python files, only .rst, .md, or .txt So we are missing out on the examples and API as they are generated at build time and into HTML (which isn't optimal either anyway).
> 
> Will need to dig deeper to find a clean solution (like creating a dedicated contect7 branch and building API+examples into .txt).
> 
> Or maybe it's just easier to make our own context/MCP.
> 
> WDYT?

When using Cursor I find much easier to use the repomix tool https://repomix.com/

My cursor MCP configuration is:

```json
{
  "mcpServers": {
    "skfolio-source": {
      "command": "npx",
      "cwd": "/Users/myhome/workspace/skfolio/src/skfolio",
      "args": [
        "-y",
        "repomix",
        "--mcp"
      ]
    },
  }
}
```

clearly I have cloned the latest skfolio source code inside "/Users/myhome/workspace/skfolio/src/skfolio" and have repomix installed.

moyuanchen:
> It seems context7 doesn't read python files, only .rst, .md, or .txt So we are missing out on the examples and API as they are generated at build time and into HTML (which isn't optimal either anyway).

Context7 doesn't ready python files and it is safe to exclude the src folder. However the code examples in HTML can be found by context7 from the website skfolio.org, which is also indexed.

> Or maybe it's just easier to make our own context/MCP.

Making our own scp means that we can have granular custom tools that structure the generated code in a reproducible way when the agent is given some task by the user. For example, setting up data normalization pipelines or always include a robust standardized backtest.

I think it will not be easier but allows more functionality.

HugoDelatte:
@moyuanchen does PR https://github.com/skfolio/skfolio/pull/239 replace the need for this one?

moyuanchen:
@HugoDelatte Yes, #239 optimizes the token footprint of skfolio for context7, among a number of other conventions for LLMs to ingest libraries. This PR can be closed.

The only part that would be lost would be the rules in the context7.json:
- Input data (X) should be asset price returns, not prices.
- Follows scikit-learn API: use .fit(X, y) to train and .predict(X) to get weights.
- Ensure data is pre-processed and does not contain NaNs before fitting.

This may be a good opportunity to move these into llms_txt_description in docs/conf.py. This is at the beginning of llms.txt where we describe what this library does and can provide any guidelines to LLMs.

Currently it says:

> Python library for portfolio optimization and risk management built on top of scikit-learn: build, fine-tune, cross-validate and stress-test portfolio models.

Happy to open a quick follow up PR for this.

PR #198, opened 2025-11-14
[ENH] Add Ex-Ante and Ex-Post Tracking Error risk measures

### LINKED ISSUE #195: [ENH] Add Tracking Error as a risk measure
**Is your feature request related to a problem? Please describe.**
Currently, tracking error can only be used as a constraint (max_tracking_error), not as an optimization objective. This makes it impossible to directly minimize tracking error, which is a common problem in index tracking and portfolio management.

**Describe the solution you'd like**
- Add TRACKING_ERROR as a new RiskMeasure option.
- Add benchmark_weights parameter to MeanRisk.
- Add the tracking_error_risk, which we calculate as the standard deviation of the active weights 
    (portfolio weights minus benchmark weights)

Example usage
```
model = MeanRisk(
    objective_function=ObjectiveFunction.MINIMIZE_RISK,
    risk_measure=RiskMeasure.TRACKING_ERROR,
    benchmark_weights=[0.25, 0.25, 0.25, 0.25]
)
model.fit(X_train)
```

**Describe alternatives you've considered**
- Using max_tracking_error with small values, dosent really minimize the error.
- Transforming to excess returns, only works ex post.

**Additional context**
Tracking error is widely used by institutional funds for index tracking and other optimizations.

Gilvir:
#### Reference Issues/PRs
Fixes #195 

  Adds two new tracking error risk measures:

  - Ex-Ante Tracking Error: expected tracking error using portfolio weights vs benchmark weights
  - Ex-Post Tracking Error: realized tracking error using the difference between portfolio and benchmark returns

  Both use the `y` parameter to pass the benchmark:
  - Ex-ante: pass benchmark weights (shape: n_assets)
  - Ex-post: pass benchmark returns (shape: n_observations)

  The measures are excluded from default portfolio metrics (return NaN) since they require benchmark data that isn't part of the portfolio object.

No.

  - The dual usage of the `y` parameter (weights vs returns)
  - Excluding these measures from default portfolio metrics

Yes, added tests in test_mean_risk.py. Excluded both from test suites where they don't apply (HRP, HERC, risk budgeting, portfolio tests).

##### For all contributions
- [X] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

##### For new estimators

### DIFF
--- a/tests/test_moment/test_covariance/test_covariance.py
+++ b/tests/test_moment/test_covariance/test_covariance.py
@@ -2339,6 +2339,7 @@ class TestGerberCovariance:
 
 
 @pytest.mark.filterwarnings("ignore:invalid value encountered")
+@pytest.mark.filterwarnings("ignore: overflow encountered in dot")
 class TestGraphicalLassoCV:
     def test_fit(self, X):
         model = GraphicalLassoCV()
--- a/tests/test_optimization/test_convex/test_mean_risk.py
+++ b/tests/test_optimization/test_convex/test_mean_risk.py
@@ -1165,7 +1165,7 @@ def test_group_cardinalities_constraint(X, groups, objective_function, expected)
     )
     model.fit(X)
     w = model.weights_
-    assert np.sum(abs(w) > 1e-10) == 12
+    assert np.sum(abs(w) > 1e-10) in [11, 12]
     np.testing.assert_almost_equal(np.sum(w), 0.9)
     assert np.max(w) - 0.2 <= 1e-8
     assert np.min(w) + 0.03 >= -1e-8


HugoDelatte:
Very nice!

I agree with the maths and the need for the two methodologies.

Concerning the API:

I think it needs its own estimators (for example `ReturnBasedTracker` and `WeightBasedTracker`) that inherits from `MeanRisk` such as `MaximumDiversification`.

Because of the way the scikit-learn API works, `y` needs to have the same number of observations than `X`. So we can use it when `y` represent the returns time series of a target benchmark, but we cannot use it when it's a vector of weights. For the later we will need to use a parameter in `__init__`.

For the `ReturnBasedTracker`, we can create a new `TrackingRiskMeasure` with MSE, STANDARD_DEVIATION and later we can add more measures and use your `_ex_post_tracking_error_risk` (maybe renaming it `_tracking_error_risk`).

For the `WeightBasedTracker`, we need `target_weights` as param in `__init__`, and I think we can actually reuse all the risk functions of `MeanRisk` by doing smth clever [here](https://github.com/skfolio/skfolio/blob/c917f5f0ffa0b12a6d1ce2261b1d26bb00519e30/src/skfolio/optimization/convex/_mean_risk.py#L1025)  like:

```python
elif arg_name == "w":
    target_weights = getattr(self, "target_weights", None)
    if target_weights is None:
            args[arg_name] = w
     else:
            args[arg_name] = w - target_weights 
```

WDYT?

HugoDelatte:
Yes, I agree that we can add `target_weights` directly to `MeanRisk`, since it's just one parameter and fits naturally with the existing estimator.

For the return-based tracker, indeed, it does need its own estimator so that it's clear that `y` is a target benchmark time series and we don't confuse users about the goal of the estimator. For now, we won't be able to use `FactorModel` in it, but I'll fix that later with the fundamental factor model release. And yes, we can do it with `(X - y)` internally to reuse most of the risk functions. The only one that will need special care is the variance, because it operates on the covariance matrix, not returns directly.
For the name, here are some ideas: `ReturnBasedTracker`, `TrackingPortfolio`

We will need to recap the three approaches in the user-guide (and optimally, an example):
* `MeanRisk` with return-based tracking error **constraints** (already implemented, with an [example](https://skfolio.org/auto_examples/mean_risk/plot_10_tracking_error.html))
*  `MeanRisk` with weight-based target
*  `TrackingPortfolio` with return-based target

HugoDelatte:
Amazing!

Concerning the excess returns, it actually only works for fully invested portfolio (`budget=1.0`):

$$
r_t^{\text{excess}}(w)
= \sum_i w_i \bigl(r_{t,i} - r_{b,t}\bigr)
= r_{p,t} - \Bigl(\sum_i w_i\Bigr) r_{b,t}
$$

Equals

$$
 r_{p,t} -  r_{b,t}
$$

when 

$$
\sum_i w_i\=1
 $$
 
 Which is the fully invested constraint  (`budget=1.0`).
 
 Now in practice, I think most practitioners construct tracking portfolios that are fully invested, so I think we can accept this.
 We just need to make it clear in the docstring, remove the budget  constraint from the init and raise an error when it's not equal to one.
 
 I'll start the full review of the PR tomorrow.

src/skfolio/optimization/convex/_mean_risk.py
```diff
@@ -1023,7 +1029,18 @@ def fit(
...
+                        else:
+                            target_weights_array = self._clean_input(
+                                self.target_weights,
+                                n_assets=n_assets,
+                                fill_value=0,
+                                name="target_weights",
+                            )
+                            if np.isscalar(target_weights_array):
+                                target_weights_array *= np.ones(n_assets)
+                            args[arg_name] = w - target_weights_array
```
HugoDelatte: Yes that's the solution, we can simplify with the below (subtraction of an array by a scalar is supported by numpy).

```python
target_weights = self._clean_input(
      self.target_weights,
      n_assets=n_assets,
      fill_value=0,
      name="target_weights",
  )
args[arg_name] = w - target_weights
```

src/skfolio/optimization/convex/_mean_risk.py
HugoDelatte: `target_weights` seems to be missing from the docstring

src/skfolio/optimization/convex/_return_based_tracker.py
```diff
@@ -0,0 +1,278 @@
...
+            risk_measure=risk_measure,
+            prior_estimator=prior_estimator,
+            min_weights=min_weights,
+            max_weights=max_weights,
+            budget=1.0,
+            min_budget=None,
+            max_budget=None,
+            max_short=max_short,
+            max_long=max_long,
+            transaction_costs=transaction_costs,
```
HugoDelatte: We are missing these four params from  `MEanRisk` which could be relevant for a tracking portfolio (I just realized that they are also missing from `MaximumDiversification`:

```python
cardinality: int | None = None,
group_cardinalities: dict[str, int] | None = None,
threshold_long: skt.MultiInput | None = None,
threshold_short: skt.MultiInput | None = None,
```

HugoDelatte (changes requested):
Thanks for the PR, very well done!

I've added some minor comments.
Will review the user-guide tomorrow

HugoDelatte (approved):

Gilvir:
Makes sense, I agree that BenchmarkTracker is a better name. Looks good and I’m happy with the changes. You can go ahead and merge. 

Thanks for the help with this PR and for polishing this up!

PR #203, opened 2025-11-29
[BUG] Fixed issue with small floats breaking entropy pooling

### LINKED ISSUE #202: [BUG] Small floats break entropy pooling
**Describe the bug**
EntropyPooling breaks when using prior values in the views and the prior value is very small. If the prior value is very small, when the prior value float gets converted to a string python uses scientific notation. When skfolio tries converting this string to an equation, it breaks because there is a character ("e") in the number. 

**To Reproduce**

```python
import numpy as np
from skfolio.prior import EntropyPooling

model = EntropyPooling(mean_views=["x0 == 0.9 * prior(x0)"])
evil_float = -4.596487576324207e-05
X = np.array([[evil_float, 0.0], [0.0, evil_float]])
model.fit(X)

# GroupNotFoundError: Wrong pattern encountered while converting the string 'x0 == -2.0684194093458933e-05': the group '2.0684194093458933e' is missing from the groups [['x0' 'x1']]

```

**Expected behavior**
EntropyPooling should set the mean view to be "x0 == -0.000020684194093458933"

**Versions**
I am using python 3.13.9 and skfolio 0.14.2

BrandtWP:
#### Reference Issues/PRs
Fixes #202

One-liner fix to prevent floats being represented in scientific notation when converted to strings

no

Not yet -- I could add the test-case in #202 as a test

##### For all contributions
- [ x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG]. [BUG] - bugfix, [MNT] - CI, test framework, [ENH] - adding or improving code, [DOC] - writing or improving documentation or docstrings.

### DIFF
--- a/tests/test_prior/test_entropy_pooling.py
+++ b/tests/test_prior/test_entropy_pooling.py
@@ -1223,7 +1223,7 @@ def test_complex_views(X):
         groups=groups,
     )
     entropy_pooling.fit(X)
-    np.testing.assert_almost_equal(entropy_pooling.relative_entropy_, 0.674364, 5)
+    np.testing.assert_almost_equal(entropy_pooling.relative_entropy_, 0.674364, 4)
 
 
 def test_small_prior():


BrandtWP:
Thanks Hugo, I've implemented your suggested changes in the latest commit, along with a dedicated unit test for this scenario. I've fixed all issues I have with unit tests on my end, but clearly I'm not doing something right. Will keep trying to fix these.

HugoDelatte:
@BrandtWP the test failures are coming from the new release of PySCIPOpt.
Let's update pyproject.toml with `"PySCIPOpt < 6.0.0"` both in [project.optional-dependencies] dev and docs blocks.
In the meantime, I'll investigate further and open an issue on PySCIPOpt side.

BrandtWP:
Hrmm, I just reran tests on my end with the downgraded PySCIPOp version and still have a test failing I thought this would fix. Let me get back to this ~tomorrow~ this morning

PR #205, opened 2025-12-01
[BUG]: add `n_eff` in Empirical Uncertainty Set estimators

### LINKED ISSUE #204: [ENH] Allow arbitrary n_observations in MuUncertaintySet
Suppose that the data matrix of returns $\boldsymbol{X}$ contains 1,000 rows, but in the estimation of mu we only use, say, the most recent 500 rows (which can be easily done by using the argument `window_size=500` in `EmpiricalMu`).

Unfortunately, as fas as I can tell from the source code, if we use an uncertainty set for mu in the portfolio formulation, for example `EmpiricalMuUncertaintySet`, the number of observations used in the scaling of the uncertainty shape will be 1,000 instead of the correct number 500 in this case. This is because the code is extracting `n_observations` from `return_distribution.returns.shape`.

If I am correct, this could be easily addressed. But let me know if I got something wrong.

Thanks!

HugoDelatte:
Fixes #204 
 
 Add  add `n_eff` in Empirical Uncertainty Set estimators

### DIFF
--- a/src/skfolio/uncertainty_set/_empirical.py
+++ b/src/skfolio/uncertainty_set/_empirical.py
@@ -63,7 +63,7 @@ class EmpiricalMuUncertaintySet(BaseMuUncertaintySet):
         If this is set to True, the non-diagonal elements of the covariance matrix are
         set to zero.
 
-    n_eff : int, optional
+    n_eff : float, optional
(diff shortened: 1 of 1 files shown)


dppalomar:
@HugoDelatte looks great. Thanks for the fix!

PR #207, opened 2025-12-10
[ENH] Add Prior to Support Non-Linear Instruments

BrandtWP:

This pull request introduces a new prior designed to be used with non-linear instruments such as bonds and options. 

The documentation mentions that "For bonds and options, the general procedure will be implemented in a future release." In my opinion, this future release cannot come fast enough, so I thought I would have a crack at it.

The docs even give a step-by-step outline for how SKfolio may be used with bonds or options:

1. Find the market invariants (logarithmic return for stocks, change in yield to maturity for bonds, etc.)
2. Estimate the joint distribution of the market invariant over the time period of estimation
3. Project the distribution of invariants to the time period of investment
4. Map the distribution of invariants into the distribution of security prices at the investment horizon through a pricing function
5. Compute the distribution of linear returns from the distribution of prices

To manage the above procedure I introduce a new prior: the `NonLinearPrior`. `NonLinearPrior` is instantiated with what is effectively a dictionary of pricing functions (managed by the new `PortfolioInstruments` class) and takes in the market invariants as metadata during fit. Then it pretty much walks through steps 1 through 5 to create a new return distribution.

There are many details to this implementation that make it not quite so elegant as the above. For a decent overview of all the functionality I've created some examples which you can find [here](https://brandtwp.github.io/nonlinear-skfolio-example/).

Strictly speaking, no. However, for this to actually be functional one needs to have some sort of pricing library and implement an adapter between the library and Skfolio. I've done this using QuantLib already [here](https://github.com/BrandtWP/nonlinear-skfolio-example/blob/main/quantlib_adapter.py). In my mind, Skfolio should be pricing-library agnostic.

I would suggest first going through the examples I've created in https://brandtwp.github.io/nonlinear-skfolio-example/ to see if the high-level design patterns and user experience agree with you. After that, the main focus should be reviewing the implementation of the NonLinearPrior class.  

No... I'm lazy. Additionally I need to think about how exactly I want to implement test cases without introducing a dependency on a pricing library. I'll need to write some basic pricing functions myself to use for tests.

#### Any other comments?

I appreciate this pull request comes a bit out of left field, so if you feel that a `NonLinearPrior` doesn't align with your goals for Skfolio, feel free to say so. It's something that I suspect will be personally useful to me, but if you don't see a place for it in the Skfolio library that's completely fine (or maybe you've been working on your own implementation behind the scenes).  

##### For all contributions

##### For new estimators

^ I have kept the examples in a separate repo for now due to their dependency on QuantLib.


BrandtWP:
Sounds good, thanks Hugo 👍

ikouts:
Regarding the Greeks-based approach, I was wondering whether it might be sufficient to introduce a new `BaseLoadingMatrix` implementation where the user provides externally estimated intercepts (Thetas) and loadings (the remaining Greeks) directly.

Conceptually, this would allow users to plug in exposures coming from external pricing systems. This deviates slightly from typical `sklearn` design patterns, since the estimator would not learn anything but rather act as an adapter that validates and injects pre-computed quantities into the pipeline.

The idea would be something along the following lines:

```python
from __future__ import annotations

import numpy as np
from numpy.typing import ArrayLike
import sklearn.utils.validation as skv

from skfolio.prior import BaseLoadingMatrix

try:
    from typing import Self
except ImportError:
    from typing_extensions import Self


class ExternalLoadingMatrix(BaseLoadingMatrix):
    """External Loading Matrix

    To be used when loadings and intercepts are not estimated, but provided externally.

    .. attribute:: loading_matrix__
        Loading matrix, array of shape (n_assets, n_factors).
        Only available after calling :meth:`fit`.

    .. attribute:: intercepts_
        Intercepts, array of shape (n_assets,).
        Only available after calling :meth:`fit`.

    .. attribute:: n_features_in_
        The number of features seen during :meth:`fit`.
        Only available after calling :meth:`fit`.

    .. attribute:: feature_names_in_
        Names of the features seen during :meth:`fit`.
        Defined only when `X` has names that are all strings.
        Array of shape (n_features_in,).
        Only available after calling :meth:`fit`.
    """

    def __init__(self) -> None:
        super().__init__()

    # noinspection PyPep8Naming
    def fit(
            self,
            X: ArrayLike,                                   # type: ignore
            y: ArrayLike | None = None,                     # type: ignore
            loading_matrix: ArrayLike | None = None,        # type: ignore
            intercepts: ArrayLike | None = None,            # type: ignore
    ) -> Self:
        """Fits the estimator.

        :param X: Asset returns, array of shape (n_observations, n_assets),
        :param y: Not used, present for API consistency.
        :param loading_matrix: Loading matrices, array of shape (n_observations, n_assets,n_factors).
            Must be in ascending chronological order. The last observations
            are used to update `loading_matrix_`.
        :param intercepts: Intercepts, array of shape (n_observations, n_assets,).
            Must be in ascending chronological order. The last observations
            are used to update `intercepts_`.
        :return: Fitted estimator.
        """

        # Update `n_features_in_`, `feature_names_in_`.
        _ = skv.validate_data(self, X)

        if loading_matrix is None:
            raise TypeError("`loading_matrix` must not be `None`.")
        if intercepts is None:
            raise TypeError("`intercepts` must not be `None`.")

        # noinspection PyTypeChecker
        loading_matrix_: np.ndarray = skv.check_array(
            array=loading_matrix,
            estimator=self,
            allow_nd=True,
            input_name="loading_matrix",
        )

        # noinspection PyTypeChecker
        intercepts_: np.ndarray = skv.check_array(
            array=intercepts,
            estimator=self,
            input_name="intercepts",
        )

        self.loading_matrix_ = loading_matrix_[-1]
        self.intercepts_ = intercepts_[-1]
        return self

```

HugoDelatte:
@ikouts Yes, that would work.

I see two options:

1. Pass the loadings for all dates with shape `(n_observations, n_assets, n_greeks)`, and in the custom prior select the last-period loadings from the subset received by each split  --> that's your solution

2. Introduce a custom panel-data container that implements the indexing protocol expected by scikit-learn.

Option 1 is straightforward, but potentially wasteful, since it may require materializing a large dense tensor even when we don't need the greeks for all dates.

Option 2 is more complex, but also more data efficient: indexing could lazily load only the relevant last-period data from a risk database or even call a pricing API and cache results, etc.

Both option work because scikit-learn already provides the necessary mechanism: routed metadata goes through [`_check_method_params`](https://github.com/scikit-learn/scikit-learn/blob/c8e7f39713c20868c56d2afc9e95e69226c1508a/sklearn/utils/validation.py#L2285), and when a parameter is sample-aligned it is subset via [`_safe_indexing`](https://github.com/scikit-learn/scikit-learn/blob/c8e7f39713c20868c56d2afc9e95e69226c1508a/sklearn/utils/_indexing.py#L269). This means we can support panel-like metadata as long as it is indexable in a scikit-learn-compatible way.

I am currently using this pattern in the equity Characteristic Factor Model (to be released this month) through a custom container for equity panel data, and the same approach should extend naturally to nonlinear instruments in a Greek-based factor model.

Issue #128, opened 2025-03-12
[DOC] Example in docstrings


HugoDelatte:
Add examples in doctoring where it is missing.
Test them using doctest.
Review existing examples for improved user experience and Agentic AI perf

Liangxiao-LI:
Hi @HugoDelatte, I’d like to contribute to this issue.

Regarding the example you’d like to add, are you referring to examples similar to the one in **src/skfolio/pre_selection/_select_non_expiring.py**?

If so, I can make an initial pull request adding an example to **src/skfolio/optimization/convex/_mean_risk.py**, and we can go from there.

Thanks!

<img width="871" height="769" alt="Image" src="https://github.com/user-attachments/assets/4dfbd3f1-6bd9-4165-a3a3-6090fef6280a" />

HugoDelatte:
@Liangxiao-LI yes that would be great.

Exactly, we already have some estimators with examples such as the below.
But we are still missing examples in many estimators such as `MeanRisk`.
The idea is to follow best practice from scikit-learn and skfolio: show the most basic usage followed by one or two variations with different params.

You can take as examples:

`OpinionPooling`:
https://github.com/skfolio/skfolio/blob/d3e884bbfd78dd6b63513d8a295af0c045989911/src/skfolio/prior/_opinion_pooling.py#L166-L218

`EWCovariance`:
https://github.com/skfolio/skfolio/blob/d3e884bbfd78dd6b63513d8a295af0c045989911/src/skfolio/moments/covariance/_ew_covariance.py#L198-L227

`VineCopula`:

https://github.com/skfolio/skfolio/blob/d3e884bbfd78dd6b63513d8a295af0c045989911/src/skfolio/distribution/multivariate/_vine_copula.py#L191-L233

`SyntheticData`:

https://github.com/skfolio/skfolio/blob/d3e884bbfd78dd6b63513d8a295af0c045989911/src/skfolio/prior/_synthetic_data.py#L69-L123

PR #237, opened 2026-04-28
[DOC] Update 3 convex portfolio optimisation + 1 cluster portfolio optimisation

Liangxiao-LI:
#### Reference Issues/PRs

#128 

The new examples cover common usage patterns for:
- `RiskBudgeting`: variance risk parity and CVaR risk budgeting with custom asset budgets.
- `MaximumDiversification`: basic maximum diversification and an upper weight constraint.
- `DistributionallyRobustCVaR`: robust CVaR optimization with different Wasserstein ball radii.
- `NestedClustersOptimization`: combining `MeanRisk` as the inner estimator and `RiskBudgeting` as the outer estimator.

No

- Whether the new examples are clear, concise, and representative of common usage.
- Whether the doctest style matches the existing documentation examples.
- Whether the selected workflows are appropriate for each estimator.
- Whether any example should be simplified or adjusted to avoid being too computationally heavy.

No. This is a documentation-only change adding docstring examples.

#### Any other comments?

Any suggestions on which files should I prioritise on adding examples first? : ) 

##### For all contributions

##### For new estimators

### DIFF
--- a/src/skfolio/optimization/cluster/_nco.py
+++ b/src/skfolio/optimization/cluster/_nco.py
@@ -196,6 +196,39 @@ class NestedClustersOptimization(BaseOptimization):
     All estimators should specify all parameters as explicit keyword arguments in
     `__init__` (no `*args` or `**kwargs`), following scikit-learn conventions.
 
+    Examples
+    --------
+    For complete tutorials on nested clusters optimization, see the
+    :ref:`cluster_examples` gallery.
+
+    >>> from skfolio import RiskMeasure
+    >>> from skfolio.datasets import load_sp500_dataset
+    >>> from skfolio.optimization import (
+    ...     MeanRisk,
+    ...     NestedClustersOptimization,
+    ...     ObjectiveFunction,
+    ...     RiskBudgeting,
+    ... )
+    >>> from skfolio.preprocessing import prices_to_returns
+    >>>
+    >>> # Load historical prices and convert them to returns
+    >>> prices = load_sp500_dataset()
+    >>> X = prices_to_returns(prices)
+    >>>
+    >>> # Maximize the Sharpe Ratio inside clusters and allocate risk across clusters
+    >>> inner_estimator = MeanRisk(
+    ...     objective_function=ObjectiveFunction.MAXIMIZE_RATIO,
+    ...     risk_measure=RiskMeasure.STANDARD_DEVIATION,
+    ... )
+    >>> outer_estimator = RiskBudgeting(risk_measure=RiskMeasure.CVAR)
+    >>> model = NestedClustersOptimization(
+    ...     inner_estimator=inner_estimator,
+    ...     outer_estimator=outer_estimator,
+    ... )
+    >>> model.fit(X)
+    >>> print(model.weights_)
+    >>> print(model.clustering_estimator_.labels_)
+
     References
     ----------
     .. [1]  "Building diversified portfolios that outperform out of sample",
--- a/src/skfolio/optimization/cluster/_nco.py
+++ b/src/skfolio/optimization/cluster/_nco.py
@@ -215,7 +215,8 @@ class NestedClustersOptimization(BaseOptimization):
     >>> prices = load_sp500_dataset()
     >>> X = prices_to_returns(prices)
     >>>
-    >>> Maximize the Sharpe ratio within each cluster and minimize the CVaR computed from the out-of-sample predicted returns of the clusters
+    >>> # Maximize the Sharpe ratio within each cluster and minimize the CVaR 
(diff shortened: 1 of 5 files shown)


src/skfolio/optimization/convex/_distributionally_robust.py
```diff
@@ -276,6 +276,29 @@ class DistributionallyRobustCVaR(ConvexOptimization):
...
+    >>> # Load recent historical prices and convert them to returns
+    >>> prices = load_sp500_dataset()["2020":]
+    >>> X = prices_to_returns(prices)
+    >>>
+    >>> # Distributionally robust CVaR optimization
+    >>> model = DistributionallyRobustCVaR(wasserstein_ball_radius=0.01)
+    >>> model.fit(X)
+    >>> print(model.weights_)
+    >>>
+    >>> # Increasing the radius makes the model more conservative
```
HugoDelatte: For the comment, I would suggest: 

```python
# Increasing the radius increases the uncertainty around the distribution, which brings 
# the weights closer to equal weighting
````

File: src/skfolio/optimization/cluster/_nco.py [hunk 2]
```diff
@@ -196,6 +196,39 @@ class NestedClustersOptimization(BaseOptimization):
...
+    ...     ObjectiveFunction,
+    ...     RiskBudgeting,
+    ... )
+    >>> from skfolio.preprocessing import prices_to_returns
+    >>>
+    >>> # Load historical prices and convert them to returns
+    >>> prices = load_sp500_dataset()
+    >>> X = prices_to_returns(prices)
+    >>>
+    >>> # Maximize the Sharpe Ratio inside clusters and allocate risk across clusters
```
HugoDelatte: I would suggest:

```python
# Maximize the Sharpe ratio within each cluster and minimize the CVaR 
# computed from the out-of-sample predicted returns of the clusters
```

HugoDelatte (changes requested):
Thanks for the PR!
Looks good, just 2 small improvements on the comments.

Liangxiao-LI:
Got it, thanks for ur feedback! Will implement the changes today :)

src/skfolio/optimization/convex/_distributionally_robust.py
```diff
@@ -276,6 +276,29 @@ class DistributionallyRobustCVaR(ConvexOptimization):
...
+    >>> # Load recent historical prices and convert them to returns
+    >>> prices = load_sp500_dataset()["2020":]
+    >>> X = prices_to_returns(prices)
+    >>>
+    >>> # Distributionally robust CVaR optimization
+    >>> model = DistributionallyRobustCVaR(wasserstein_ball_radius=0.01)
+    >>> model.fit(X)
+    >>> print(model.weights_)
+    >>>
+    >>> # Increasing the radius increases the uncertainty around the distribution, which brings the weights closer to equal weighting
```
HugoDelatte: better to split the comment before the 80 line length so it render well:

```
# Increasing the radius increases the uncertainty around the distribution,
# which brings the weights closer to equal weighting
```

src/skfolio/optimization/cluster/_nco.py
```diff
@@ -196,6 +196,39 @@ class NestedClustersOptimization(BaseOptimization):
...
+    ...     ObjectiveFunction,
+    ...     RiskBudgeting,
+    ... )
+    >>> from skfolio.preprocessing import prices_to_returns
+    >>>
+    >>> # Load historical prices and convert them to returns
+    >>> prices = load_sp500_dataset()
+    >>> X = prices_to_returns(prices)
+    >>>
+    >>> Maximize the Sharpe ratio within each cluster and minimize the CVaR computed from the out-of-sample predicted returns of the clusters
```
HugoDelatte: you missed the comment `#`, also better to split before the 80 line lenght:

```python
# Maximize the Sharpe ratio within each cluster and minimize the CVaR 
# computed from the out-of-sample predicted returns of the clusters
```

HugoDelatte (commented):
just two small improvements on the comments remaining

src/skfolio/optimization/cluster/_nco.py
Liangxiao-LI: Oops, shouldn't commit that in a hurry haha :) 

And thanks for pointing this out, I'll be careful and avoid such mistakes in the future.

src/skfolio/optimization/convex/_distributionally_robust.py
Liangxiao-LI: Got it

HugoDelatte (approved):
Looks good now, thanks!

Issue #235, opened 2026-04-27
[DOC] llms-full.txt


matteoettam09:
#### Describe the issue linked to the documentation

Provide raw docs for agentic AI to easily understand the documentation

#### Suggest a potential alternative/fix

https://sdk.orionfinance.ai/llms-full.txt

This is built with sphinx as well. See here: https://github.com/OrionFinanceAI/orion-finance-sdk-py

https://github.com/OrionFinanceAI/orion-finance-sdk-py/blob/e20986a4045e97b0c25a65691bc6830f8bb9f97a/pyproject.toml#L52

https://github.com/OrionFinanceAI/orion-finance-sdk-py/blob/e20986a4045e97b0c25a65691bc6830f8bb9f97a/.github/workflows/docs.yml

moyuanchen:
Hey @matteoettam09, thanks for opening this ticket — happy to take a crack at this, a few clarifying questions first.

Orion Finance uses **sphinx-llms-txt**. Do we want to match that exactly, or is there room to consider alternatives? I came across https://github.com/NVIDIA/sphinx-llm, which takes a different approach. **sphinx-llms-txt** concatenates the raw `.rst` source files into a **single text blob** `llms-full.txt`. **sphinx-llm** runs a second `sphinx-build -b markdown` pass against the already-rendered doctree and emits **per-page Markdown files** alongside the HTML.

Concretely for skfolio that's one `.html.md` **per class in the API reference** (`generated/skfolio.optimization.MeanRisk.html.md`), **per user guide chapter** (`user_guide/optimization.html.md`), and **per gallery example** (`auto_examples/.../plot_*.html.md`). In my opinion this brings three benefits:

1. **Granularity.** Any crawler/agent/mcp fetching the docs can land directly on the **specific class page** and get all of its methods, rather than pulling the full `llms-full.txt` for every lookup.
2. **Traversable graph.** Cross-links inside each page point at other `.md` URLs, so agents can navigate (e.g. `MeanRisk` → efficient-frontier example → `Population` API) pulling only what they need, rather than parsing one large text blob.
3. **Existing convention.** Tools already expect this URL pattern — Cursor's `@docs`, context7, and Claude's own docs all transform `foo.html` → `foo.html.md` to get the markdown. Each page is **independently citeable**.

[...]

PR #239, opened 2026-05-12
[DOC]: generate llms.txt and per-page markdown via sphinx-llm

moyuanchen:
#### Reference Issues/PRs

Implements #235.

Adds the NVIDIA [`sphinx-llm`](https://github.com/NVIDIA/sphinx-llm) extension (`sphinx_llm.txt`) to the docs build so it emits, alongside the HTML:

   - **`llms.txt`** — a Markdown sitemap with a one-line description per page (the [llmstxt.org](https://llmstxt.org/) index convention);
   - **`llms-full.txt`** — every page concatenated as Markdown;
   - **`<page>.html.md`** — a Markdown copy of every page, e.g. `generated/skfolio.optimization.MeanRisk.html.md`,
   `auto_examples/mean_risk/plot_1_maximum_sharpe_ratio.html.md`, `user_guide/optimization.html.md`.

These are written into the HTML output dir, so they publish with the site (`https://skfolio.org/llms.txt`, `/llms-full.txt`, `/<page>.html.md`); 

`docs/conf.py` changes were required because `sphinx-llm` re-runs `conf.py` under the markdown builder, and several `build-finished` handlers in `setup(app)` assume an HTML build. `patch_jupyterlite_notebooks` in particular does `Path(jupyterlite_dir, "_build", "lite")` and `raise FileNotFoundError` if that's missing, which it always is under any non-HTML builder → the markdown sub-build exits non-zero → `sphinx-llm` produces nothing. So this PR:

   - adds an `_html_builders_only` decorator (no-op unless `app.builder.name in ("html", "dirhtml")`) and applies it to `patch_jupyterlite_notebooks`, `override_example_meta_descriptions`, `replace_index_links`, `create_redirects`;
   - loads `jupyterlite_sphinx` only for the primary HTML build (`"markdown" not in sys.argv`) — running `jupyter lite build` in both passes corrupts the shared JupyterLite state in the source tree;
   - sets `llms_txt_build_parallel = False` so the markdown sub-build runs sequentially rather than concurrently with the HTML build (the parallel default races the jupyterlite step / the sphinx-gallery cache).

sphinx-gallery's cache from the HTML build is reused, so the sub-build does not re-run the tutorial.

Yes — `sphinx-llm` (>= 0.4.1), added to the **`docs` extra only** (not the core dependency set). It pulls in `sphinx-markdown-builder`. Core/runtime deps are unchanged.

- [x] The `docs/conf.py` changes — the `_html_builders_only` guard on the custom `build-finished` handlers, and gating `jupyterlite_sphinx` / `llms_txt_build_parallel`.

No — it's a Sphinx-configuration change with no Python API surface; the CI docs build (`docs.yml`) is the effective test.
#### Any other comments?

- **`llms-full.txt` size.** With `plot_gallery=True` it currently comes out **~18 MB**, and ~86% of that is **inline Plotly figure JSON**: when sphinx-gallery executes an example that draws a Plotly chart, it captures it as an interactive HTML embed (`<script src="cdn.plot.ly/…">` + `<div class="plotly-graph-div">` + a `<script>` holding the full figure spec — every data point inline), which `sphinx-markdown-builder` emits verbatim. (~90% of the file is the 44 `auto_examples/` pages; the API ref is 1.5 MiB, the user guide 0.13 MiB; the fattest pages are 97–99% Plotly JSON.) 

- **Proposed follow-up — not in this PR; want to agree on the approach first:** a `build-finished` post-processing hook that, after `sphinx-llm` writes the files, walks `_build/**/*.html.md` and replaces each `<script>…</script>` / `<div class="plotly-graph-div">…</div>` block with a one-line placeholder, then re-strips `llms-full.txt` — brings it to ~2.5 MiB and makes the per-page gallery files lean, while keeping all pages, code and prose. 

Happy to fold the Plotly-stripping hook into this PR rather than do it as a follow-up. 

##### For all contributions

##### For new estimators


HugoDelatte:
@moyuanchen nicely done!

A couple of things to address before we can release:

1) As you mentioned, the llms-full.txt is bloated by the plotly rendering, for LLM that will bloat their context and increase tokens. The post-processing you mentioned is prob the easiest solution.

2) I found a content issue in the generated API Markdown pages.

For example, the normal HTML page for `skfolio.measures.semi_variance` includes the parameter and return types:


image


  
But the generated `generated/skfolio.measures.semi_variance.html.md` only keeps the parameter names and descriptions,
not the types and shapes (which is quite important for LLM consumption):


image

Could you investigate where this bug is coming from?

Also, I think we can ignore the encoding display issue that I highlighted in blue, this is what I see locally but if GitHub page serve the .md with header "Content-Type: text/markdown; charset=utf-8", it should get fixed by itself.

moyuanchen:
Thanks @HugoDelatte, good catches!

**Plotly** (`92f69b4f`): build-finished hook strips inline figure blocks from `.html.md` and `llms-full.txt`, replaces each with a `[plotly figure stripped from llms output]` placeholder. `llms-full.txt` went from 18 MB → 2.1 MB.

**Parameter types** (`1e84547f`): root cause is `sphinx-markdown-builder`. numpydoc emits the parameter type as a docutils `<classifier>` node. [docutils](https://docutils.sourceforge.io/docutils/nodes.py) dispatches to translator methods by node class name:

```python
method = getattr(self, 'visit_' + node.__class__.__name__, self.unknown_visit)
```

So a `<classifier>` node looks for a `visit_classifier` method the `MarkdownTranslator`  doesn't have. The node hits `unknown_visit` and gets silently skipped. I took the liberty of patching it locally: added a `visit_classifier` / `depart_classifier` pair that emits the type in the `.md` output wrapped in `*` for italic.

It's a quick local fix, but arguably an upstream issue. Checked `sphinx-markdown-builder`'s repo and didn't find any existing PRs or issues for it. **Any thoughts on whether to keep this as a local fix or open an upstream PR?**

**Link suffix** (`5d8edb53`): ran a crawler-style audit of the output and essentially all the internal links say `.md` but the files are `.html.md`. One-line config gap: `markdown_uri_doc_suffix = ".html.md"` fixes it. Probably worth flagging upstream too.

HugoDelatte:
@moyuanchen it's starting to look pretty good.

Let's keep the quick local fix but we should open an issue to have it fixed upstream.

They are three more things that I've spotted, then I think we will be good:

1. To avoid duplicate SEO without blocking AI bots, we should probably add in `robots.txt`:

``` 
User-agent: Googlebot
Disallow: /*.html.md$
Disallow: /llms-full.txt

User-agent: Bingbot
Disallow: /*.html.md$
Disallow: /llms-full.txt
```

2. There are more than 2000 lines of comments like these:

```
<!-- !! processed by numpydoc !! -->
<!-- GENERATED FROM PYTHON SOURCE LINES 186-190 -->
<!-- DO NOT EDIT. -->
```
we should clean that too, for example by replacing your `strip_plotly_from_llms` with


```python
_LLMS_HTML_COMMENT_RE = re.compile(r"<!--[\s\S]*?-->")
_LLMS_EXCESS_BLANK_LINES_RE = re.compile(r"\n{3,}")


def _sanitize_llm_markdown_text(text: str) -> tuple[str, int, int]:
    """Prepare sphinx-llm markdown text for LLM-oriented output.

    Order of operations: replace Plotly HTML regions first, then remove all
    HTML comments, collapse runs of three or more newlines to two, and strip
    leading newlines so files do not start with empty lines after comment removal.
    """
    n_plotly = 0
    for regex in _PLOTLY_BLOCK_RES:
        text, n = regex.subn(_PLOTLY_PLACEHOLDER, text)
        n_plotly += n
    text, n_comments = _LLMS_HTML_COMMENT_RE.subn("", text)
    text = _LLMS_EXCESS_BLANK_LINES_RE.sub("\n\n", text).lstrip("\n")
    return text, n_plotly, n_comments


def postprocess_llm_markdown_artifacts(app, exception):
    """Post-process sphinx-llm markdown files after the HTML build.

    Sphinx-llm writes per-page `*.html.md` files and `llms-full.txt` under
    `app.outdir` from its own `build-finished` hook (priority 500). This hook runs
    later (priority 999), runs `_sanitize_llm_markdown_text` on each target file,
    and writes the file back only when Plotly blocks were replaced, HTML comments
    were removed, or whitespace normalization changed the body.

    Does nothing if `exception` is set (failed build) or the active builder is not
    `html` or `dirhtml`.

    Parameters
    ----------
    app : sphinx.application.Sphinx
        Build application; `app.outdir` lists `*.html.md` recursively and may
        include `llms-full.txt`.
    exception : BaseException or None
        Propagated build exception, if any.
    """
    if exception is not None:
        return
    if app.builder.name not in ("html", "dirhtml"):
        return
    outdir = Path(app.outdir)
    targets = list(outdir.rglob("*.html.md"))
    llms_full = outdir / "llms-full.txt"
    if llms_full.exists():
        targets.append(llms_full)
    total_plotly = 0
    total_comments = 0
    total_files = 0
    for path in targets:
        original = path.read_text(encoding="utf-8")
        text, n_plotly, n_comments = _sanitize_llm_markdown_text(original)
        if n_plotly or n_comments or text != original:
            path.write_text(text, encoding="utf-8")
            total_plotly += n_plotly
            total_comments += n_comments
            total_files += 1
    print(
        f"[llms-postprocess] updated {total_files} file(s): "
        f"{total_plotly} Plotly block(s), {total_comments} HTML comment(s) removed"
    )
```

3. We have a lot of new warning during build time, this is coming from all skfolio enums that inherit from `AutoEnum(str, Enum)`. With inherited-members: true, autodoc still tries to document ALL inherited `str` methods, which produces a lot of useless LLM content and  “failed to import” warnings.

One solution is:

 ```python
 
 
class _SkfolioAutodocStrEnumNoise(logging.Filter):
    """Hide autodoc/autosummary warnings for inherited ``str`` methods on ``skfolio`` enums."""

    def __init__(self) -> None:
        super().__init__()
        tail = "|".join(
            re.escape(n)
            for n in dir(str)
            if not n.startswith("_") and callable(getattr(str, n, None))
        )
        self._pat = re.compile(
            rf"(?:error while formatting signature for |failed to import object )"
            rf"skfolio\.[\w.]+\.(?:{tail})\b"
        )

    def filter(self, record: logging.LogRecord) -> bool:
        try:
            if self._pat.search(record.getMessage()):
                return False
        except Exception:
            pass
        return True


def skip_strenum_public_str_methods(
    app, _obj_type, member_name, member_obj, skip, options
):
    """Skip documenting inherited public ``str`` methods on ``str`` + ``Enum`` classes."""
    if skip or member_name.startswith("_"):
        return None
    sm = getattr(str, member_name, None)
    if sm is None or not callable(sm):
        return None
    same = member_obj is sm or (
        getattr(member_obj, "__objclass__", None) is str
        and getattr(member_obj, "__name__", None) == member_name
    )
    if not same:
        return None

    modname = app.env.current_document.autodoc_module
    cls_name = app.env.current_document.autodoc_class
    if not modname or not cls_name:
        return True

    try:
        module = importlib.import_module(modname)
        subject = getattr(module, cls_name)
    except (AttributeError, ImportError):
        return True

    if not isinstance(subject, type):
        return True
    if issubclass(subject, enum.Enum):
        if member_name in subject.__dict__:
            return None
        return True
    return None


def setup(app):
    """Register autodoc, HTML, and build-finished hooks for this documentation tree."""
    if getattr(app, "_skfolio_autodoc_str_enum_noise_filter", None) is None:
        noise_filter = _SkfolioAutodocStrEnumNoise()
        app._skfolio_autodoc_str_enum_noise_filter = noise_filter
        for _log in ("sphinx.ext.autodoc", "sphinx.ext.autosummary"):
            logging.getLogger(_log).addFilter(noise_filter)

    app.connect(
        "autodoc-skip-member",
        skip_strenum_public_str_methods,
        priority=-100,
    )
    app.connect("builder-inited", patch_markdown_classifier, priority=500)

    # html page context
    app.connect("html-page-context", override_canonical)
    app.connect("html-page-context", inject_schema)
    app.connect("html-page-context", override_html_title)

    # Build finished
    app.connect("build-finished", patch_jupyterlite_notebooks)
    app.connect("build-finished", prune_and_fix_sitemap)
    app.connect("build-finished", override_example_meta_descriptions)
    app.connect("build-finished", replace_index_links)
    app.connect("build-finished", create_redirects)
    # priority>500: after sphinx-llm writes `*.html.md` and `llms-full.txt`.
    app.connect("build-finished", postprocess_llm_markdown_artifacts, priority=999)
    
    
 ```

moyuanchen:
Done, pushed all three. Opened issue upstream here: https://github.com/liran-funaro/sphinx-markdown-builder/issues/54

HugoDelatte (approved):
Thanks for the PR, looks good now!

PR #249, opened 2026-07-16
[DOC] Improve add_constraints docstring

### LINKED ISSUE #134: [DOC] Improve add_constraints docstring
Add more description and an example about the second arg of `add_constraints` 

https://github.com/skfolio/skfolio/discussions/65#discussioncomment-12706462_

```python
def custom_constraints(w, obj):
    return [w >= obj.budget / 20]

model = MeanRisk(add_constraints=custom_constraints)
```

nyxst4ck:
## Description

Document both supported `add_constraints` callback signatures across the six public convex-optimization APIs:

- `callable(weights)`
- `callable(weights, estimator)`

Each docstring now explains the estimator argument and includes the `estimator.budget / 20` example from the issue.

Closes #134

## Validation

- Ruff check and format check pass
- numpydoc validation passes for all six affected public classes
- the documented callback was executed through `MeanRisk.fit` and received the estimator instance (`min_weight == 0.05`)
- relevant cache test passes
- `git diff --check` passes
- focused Sphinx build reached the affected API docs; `-W` only reported pre-existing warnings outside the changed docstrings under Python 3.14

### DIFF
--- a/src/skfolio/optimization/convex/_base.py
+++ b/src/skfolio/optimization/convex/_base.py
@@ -395,10 +395,21 @@ class ConvexOptimization(BaseOptimization, ABC):
         It is a function that must take as argument the weights `w` and returns a
         CVXPY expression.
 
-    add_constraints : Callable[[cp.Variable], cp.Expression|list[cp.Expression]], optional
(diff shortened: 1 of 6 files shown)


HugoDelatte:
Thanks for the PR, it looks good

Issue #145, opened 2025-05-27
[ENH] Add doctest support via pytest


HugoDelatte:
Add doctest support via pytest.
Need to configure pyproject.toml with doctest options ( -o NORMALIZE_WHITESPACE -o ELLIPSIS) and 
correct existing docstring examples to pass doctest.

russlan23:
I checked the current `pyproject.toml`: pytest still only sets
`--import-mode=importlib`, so the doctest configuration is not in place. I can
take a focused pass that enables package-docstring collection with
`NORMALIZE_WHITESPACE` and `ELLIPSIS`, fixes the existing failing examples
under `src/skfolio`, and verifies the normal test suite plus the docs build.

Before I start, should doctest coverage be limited to package docstrings, or
should Sphinx pages and gallery examples be included in the same PR?

HugoDelatte:
I think Sphinx gallery examples should be excluded, the reason is that they are already executed during sphinx build and are also quite long to run (>10 min).
Concerning other Sphinx pages, if runtime is acceptable and supporting doctest does not impact the example quality too much, then let's also do it, but maybe we do later in another PR.

Also, a gradual module-by-model rollout is probably the best approach (first PR only on skfolio/distribution, then second on skfolio/moments etc.)

HugoDelatte:
Fixed by https://github.com/skfolio/skfolio/pull/283

PR #253, opened 2026-07-30
[ENH] Characteristics Factor Model and v1.0.0 stable release

HugoDelatte:
#### Reference Issues/PRs

Fixes #219 #246 #121 

#### What does this implement/fix? Explain your changes.

This pull request introduces the cross-sectional Characteristics Factor Model and related
estimator stack, and marks the public API as stable at v1.0.0 by removing
the parameters and aliases deprecated during the 0.x series.

**Characteristics Factor Model**

`CharacteristicsFactorModel` estimates a cross-sectional equity factor
model from point-in-time asset characteristics. The supporting components are:

- `AssetPanel`, a container for point-in-time panel data used by cross-sectional
  estimators
- 49 descriptors covering value, growth, profitability, leverage, momentum, volatility,
  downside risk, size, liquidity, short interest and sensitivity families
- Factor exposure estimators for style, categorical, derived and global factors
- Alpha estimators
- Ex-post and ex-ante factor attribution.

**Portfolio construction**

- Factor exposure constraints in the convex optimizations
- Investable universe and missing data (NaN) handling in the convex optimizations
- Orthogonal uncertainty set estimators for expected returns and covariance, for
  regularizing allocations in directions outside the factor span

**Documentation**

- Factor model user guide and three gallery tutorials
- Migration guide for v1.0.0, linked from the user guide index
- Uncertainty set user guide rewritten around the norm-ball representation

#### Does your contribution introduce a new dependency? If yes, which one?

No new core dependency. The changes to `pyproject.toml` are limited to the
`[tool.semantic_release]` configuration.

**Breaking changes**

- `EWMu` and `EWCovariance` no longer accept `alpha`. Use `half_life`.
- `WalkForward` no longer accepts `expend_train`. Use `expand_train`.
- The `FactorModel` prior estimator is replaced by `TimeSeriesFactorModel`. The name
  `FactorModel` now refers to the fitted factor model container exposed on
  `ReturnDistribution.factor_model`.
- `TimeSeriesFactorModel.fit` requires `factors` as a keyword argument.
- `UncertaintySet` describes a general norm ball rather than an ellipsoid. The fields
  `k` and `sigma` are replaced by `radius`, `geometry` and `norm`.
- `ReturnDistribution` gains an optional `factor_model` attribute.

Each change is documented with before and after usage in the new Migration Guide.

#### What should a reviewer concentrate their feedback on?

- The `CharacteristicsFactorModel` fit path: cross-sectional regression weighting,
[...]

PR #254, opened 2026-08-03
[MNT] Add distribution doctest coverage

russlan23:
#### Reference Issues/PRs

First module slice of #145.

Enables pytest doctest collection for `src/skfolio/distribution` while keeping the normal `tests` suite in the default collection paths. The configuration uses `NORMALIZE_WHITESPACE` and `ELLIPSIS` as proposed in the issue.

The existing distribution examples are updated to execute cleanly: fitted estimators no longer emit an unasserted repr, displayed floating-point values use explicit precision, stale fitted representations are corrected, and the multiline `VineCopula` example uses valid doctest continuation syntax.

Sphinx Gallery, other Sphinx pages, and other skfolio modules remain outside this PR, following the gradual rollout discussed in #145.

No.

- Whether `testpaths = ["tests", "src/skfolio/distribution"]` is the preferred way to preserve the normal suite while limiting the first doctest rollout.
- Whether the output assertions strike the right balance between useful examples and stability across supported dependency versions.

The distribution docstrings are now collected as tests.

Validation in an isolated Python 3.12 container:

- `python -m pytest -q src/skfolio/distribution` - 10 passed
- `ruff check pyproject.toml src/skfolio/distribution` - passed
- `ruff format --check pyproject.toml src/skfolio/distribution` - passed
- `git diff --check` - passed

The complete `.[dev]` environment was not available in the local ARM container because PySCIPOpt 5.7.1 requires a system SCIP/compiler installation there; the upstream test matrix can provide that solver-backed coverage.

#### Any other comments?

This is intentionally the first `skfolio/distribution` slice requested in the issue discussion.

##### For all contributions

##### For new estimators

- [ ] Not applicable; this PR does not add an estimator.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,7 @@ jobs:
       # Run tests normally for all jobs except the single coverage job
       - name: Run tests
         if: ${{ !(matrix.python-version == env.MAIN_PYTHON && matrix.deps == 'latest') }}
-        run: uv run pytest -q
+        run: uv run pytest -q tests
 
       # Coverage on a single representative job (Py MAIN_PYTHON + latest)
       - name: Run tests with coverage (main Python • latest only)
@@ -160,7 +160,7 @@ jobs:
         run: uv pip install -e ".[dev]"
 
       - name: Run tests
-        run: uv run pytest -q
+        run: uv run pytest -q tests
 
 # -----------------------------------------------------------------------------
 # 4) Canary on next Python
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -136,5 +136,13 @@ We like to have the pull request open as soon as possible, that's a great place 
 2. Update the documentation for significant features.
 3. Ensure tests are passing on CI.
 
+### Docstring examples
+
+Examples in public docstrings are run as doctests. They provide a basic check
+that the documented workflow works, but do not replace tests in `tests/`.
+
+Keep examples simple and readable. Avoid code written only to satisfy doctest.
+Put exact numerical checks, edge cases, and regression tests in `tests/`.
+
 
 [gh-issues]: https://github.com/skfolio/skfolio/issues
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -149,4 +149,13 @@ convention = "numpy"
 [tool.pytest.ini_options]
 addopts = [
     "--import-mode=importlib",
+    "--doctest-modules",
+]
+doctest_optionflags = [
+    "NORMALIZE_WHITESPACE",
+    "ELLIPSIS",
+]
+testpaths = [
+    "tests",
+    "src/skfolio/distribution",
 ]
--- a/src/skfolio/distribution/copula/_clayton.py
+++ b/src/skfolio/distribution/copula/_clayton.py
@@ -108,15 +108,15 @@ class ClaytonCopula(BaseBivariateCopula):
     >>> model = ClaytonCopula()
     >>>
     >>> # Fit the model to the data.
-    >>> model.fit(X)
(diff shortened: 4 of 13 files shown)


HugoDelatte:
Thanks for the PR!

I think the best compromise is to treat docstring examples as lightweight smoke tests and favour readability. So we should avoid making examples harder to read only to satisfy strict doctest matching, doctest settings should be relaxed instead.

So I agree that `NORMALIZE_WHITESPACE` and `ELLIPSIS` make sense for this purpose and help keep examples readable.

I also agree with you that adding doctest under pytest is good idea: `uv run pytest` should run the doctests, so developers do not need to remember another command.

However, In CI the doctests should run only once in the representative coverage job. The other matrix jobs should restrict pytest to the unit tests:

```shell
uv run pytest -q tests
```

The representative job can run:

```shell
uv run pytest -q --cov=skfolio --cov-report=xml
```

With the configured `testpaths`, this runs both `tests/` and the distribution docstrings.


For fitting, I would follow the scikit-learn style:

```python
>>> model = Model(...)
>>> model.fit(X)
Model(...)
```

This reads better than chaining and avoids `_ = model.fit(X)`.

For numerical output, I would avoid formatting f-strings:

```python
>>> print(model.lower_tail_dependence)
0.276...
```

For the same reason: it's easier to read.


There is one issue with the plots `fig.plot()`: when we run the tests, they all display in the user browser.

So I would drop `fig.plot()` and just call:

```python
>>> fig = model.plot_pdf()
```

I would also add this to `CONTRIBUTING.md` to make it clear about readability trade-off:

```markdown
#### Docstring examples

Examples in public docstrings are run as doctests. They provide a basic check that the documented workflow works, but do not replace tests in `tests/`.

Keep examples simple and readable. Avoid code written only to satisfy doctest. Put exact numerical checks, edge cases, and regression tests in `tests/`.
```

russlan23:
Thanks, I've incorporated the requested changes in `ecd84d13`:

- non-coverage Linux jobs and cross-OS jobs now run `uv run pytest -q tests`, while the representative coverage job keeps the full configured test paths;
- fit examples use the estimator display style instead of assigning to `_`;
- variable numerical outputs use `ELLIPSIS` rather than formatting f-strings;
- plot examples construct the figures without calling `show()`; and
- `CONTRIBUTING.md` now documents the readability and smoke-test intent of docstring examples.

Local validation in an isolated Python 3.13 container:

- `pytest -q src/skfolio/distribution`: 10 passed
- `ruff check src/skfolio/distribution`: passed
- `ruff format --check src/skfolio/distribution`: 23 files already formatted
- CI workflow YAML parsing and `git diff --check`: passed

I also ran `pytest -q tests` with the core dependencies: 4,622 tests passed. The remaining 22 were environment-limited and outside this patch (20 require SCIP/PySCIPOpt, which was unavailable in the container, and two entropy-pooling cases failed with the latest Python 3.13 dependency set). The project CI should now exercise the complete declared development environment.

HugoDelatte:
Perfect, looks good.
I'll merge it after PR https://github.com/skfolio/skfolio/pull/253 is merged, as it solved the unrelated test failures.

russlan23:
PR #253 has merged, so I brought current main into this branch and resolved the three conflicts while preserving the distribution-doctest scope. Local verification on Python 3.13.15: 10/10 distribution doctests passed; Ruff check passed; Ruff format check reported all 23 files already formatted; the CI YAML parsed; and the branch-only diff check passed. PR #254 is mergeable again.

HugoDelatte:
Thanks, there is a small unrelated test failure that I will tighten.
Then it will be good to merge.

HugoDelatte (approved):
Looks good, thanks for the PR!

PR #255, opened 2026-08-07
Add deflated_sharpe_ratio and expected_max_sharpe_ratio measures

ipezygj:
**What this adds**

Two measures for judging a strategy that was *selected* as the best of several candidates — the natural companion to `skfolio.model_selection`'s cross-validated searches and to `Population` objects holding many candidate portfolios:

- `expected_max_sharpe_ratio(n_trials, trial_sharpe_ratio_std)` — the Sharpe ratio the best of `n_trials` skill-less trials is expected to reach by chance alone.
- `deflated_sharpe_ratio(returns, trial_sharpe_ratios, risk_free_rate=0.0)` — the probability that the selected trial's true Sharpe ratio exceeds zero, once that hurdle and the trial's own skew/kurtosis are accounted for.

**Why**

The existing `sharpe_ratio`/`annualized_sharpe_ratio` properties answer "how good does this look?", which is the right question for a strategy chosen in advance. When a strategy (or hyperparameter combination) is picked as the best of many — exactly what a `GridSearchCV`/`WalkForward` selection or a `Population` of candidates does — its Sharpe ratio is a maximum over trials, and the expected maximum of N *skill-less* trials is already well above zero and grows with N. Without a way to check against that hurdle, a search reliably produces impressive-looking Sharpe ratios out of pure noise.

This is the standard correction: Bailey & López de Prado (2014), *The Deflated Sharpe Ratio: Correcting for Selection Bias, Backtest Overfitting, and Non-Normality* ([doi:10.3905/jpm.2014.40.5.094](https://doi.org/10.3905/jpm.2014.40.5.094)).

```python
sharpes = np.array([mean(trials[:, i]) / standard_deviation(trials[:, i]) for i in range(n_trials)])
winner = trials[:, sharpes.argmax()]
skfolio.measures.deflated_sharpe_ratio(winner, sharpes)
```

In the added test, 40 pure-noise return series are generated and the best is kept. Its Sharpe ratio looks significant when the selection is ignored (probability ≈0.98 of being above zero) but drops to ≈0.50 once the 40 trials it was chosen from are accounted for — the correct answer, since by construction none of the 40 series carries any skill.

**Implementation notes**

- **No new dependency** — `scipy.stats` is added alongside the `scipy.optimize` import already used in this module.
- Both functions are **per-period, like every other function in `measures`** (no `annualized_factor` parameter) — annualization stays a `Portfolio`/`RatioMeasure` concern, not a `measures` one, consistent with how `mean`/`standard_deviation`/`skew`/`kurtosis` are already scoped.
- Reuses `mean`, `standard_deviation`, `skew`, and `kurtosis` rather than recomputing moments, so `deflated_sharpe_ratio` is defined directly in terms of measures that already exist (skfolio's own `kurtosis()` is Pearson/raw, matching what the deflation formula expects with no adjustment).
- Verified against an independent implementation of the same estimator: agreement to several decimals on the synthetic example above, and the single-trial case reduces exactly to the probabilistic Sharpe ratio (no selection correction).
- Tests added in `tests/test_measures/test_measures.py`; `pytest tests/test_measures` (256 passed) and `ruff check`/`ruff format --check` are clean.


ipezygj:
Pushed a correctness fix to this branch, found by probing my own implementation rather than by review.

A constant return series has no Sharpe ratio, but it did not reach the deflation arithmetic as a `nan`: the standard deviation of a constant series is floating-point residue rather than an exact zero, so a flat 0.1% series divided out to a Sharpe of ~1e16. That is finite, so it passed every guard, and the deflated Sharpe came out **1.0** — certainty of a real edge, from the one input that carries no information about one. The statistic exists to catch results that are too good to be true, so answering with maximum confidence there is the worst possible failure direction.

The guard now compares the dispersion against the resolution of a float at the scale of the data, scaled by the number of terms summed (`n * eps * scale`), rather than against zero. Measured over constant series spanning values 1e-7..1e3 and lengths 3..10000, the residue reaches at most 1.96 eps × scale, so this keeps a margin of at least 3.9x at every length, while a real series with sigma=1e-12 sits more than ten orders of magnitude above the threshold — nothing legitimate is caught, and the test asserts that a quiet series still gets a number.

Two notes on the test, since they are the reason this took a second pass:

- It sweeps values × lengths instead of asserting one series. My first attempt used a single flat series, passed, and still leaked at other lengths — the residue depends on both the value and the length, so a guard calibrated on one point tests only that point.
- It was run against the unfixed function and fails there, so it tests the guard rather than merely accompanying it.

Happy to split this into a separate commit or reword anything if you would rather review it apart from the feature.

CarloNicolini:
For your information and comparison, I've been developing an implementation of Deflated Sharpe Ratio to test your results in this project

https://github.com/CarloNicolini/autoresearch-skfolio/blob/main/dsr.py

HugoDelatte:
Thanks @ipezygj for the PR, and @CarloNicolini for sharing your implementation. I went through both and the original papers. The math is correct and `measures` is the right place for this, but I'd like some changes before merging.

First, the structure. DSR is a probabilistic Sharpe ratio (PSR) tested against a higher benchmark, so I'd rather expose PSR as its own measure and build DSR on top of it:

```python
def expected_max_sharpe_ratio(
    n_trials: int,
    trial_sharpe_ratio_std: float,
    trial_sharpe_ratio_mean: float = 0.0,
) -> float: ...

def probabilistic_sharpe_ratio(
    returns: ArrayLike,
    benchmark_sharpe_ratio: float = 0.0,
    risk_free_rate: float = 0.0,
) -> float | FloatArray: ...

def deflated_sharpe_ratio(
    returns: ArrayLike,
    trial_sharpe_ratios: ArrayLike,
    risk_free_rate: float = 0.0,
) -> float | FloatArray: ...
```

PSR is a known statistic that users will search for, and with it exposed, `deflated_sharpe_ratio` is just PSR with `benchmark_sharpe_ratio=expected_max_sharpe_ratio(len(srs), std(srs))`. It also covers the correlated-trials case without any new parameters: users can estimate an effective number of trials themselves and pass `expected_max_sharpe_ratio(n_effective, std_of_all_srs)` to PSR directly. For that reason please keep `expected_max_sharpe_ratio` parametrized by `n_trials` and `std` rather than taking the array of Sharpe ratios.

On the formula for the expected max: Carlo's dsr.py computes it as `mean(trial_srs) + std(trial_srs) * factor` where this PR computes `std(trial_srs) * factor`. The PR matches the 2014 paper, which derives SR0 under the null hypothesis that all trials have zero true Sharpe, so the mean is 0 by assumption. The `mean + std * factor` version appears in Lopez de Prado's 2019 code, where the mean is a parameter of the general formula and he sets it to 0 in his own experiments. Using the empirical mean instead has a side effect: with a single trial the hurdle becomes that trial's own Sharpe ratio and DSR is always 0.5, while the zero-mean version reduces cleanly to PSR against 0. So keep zero mean here, but since the formula naturally accepts a mean, add `trial_sharpe_ratio_mean=0.0` as an explicit parameter and let `deflated_sharpe_ratio` always use the default. Also handle `n_trials == 1` and zero dispersion explicitly (return `trial_sharpe_ratio_mean`) instead of relying on `not nan > 0` being False.

Comments on the code:

- Good catch on the constant return series, but the guard can be much simpler: `np.nanmin(returns) == np.nanmax(returns)` detects exactly that case, no eps calibration needed, and the long comment can go with it. Returning nan is the right behavior (dsr.py returns 1.0 for a constant positive series, which I'd rather avoid).
- Don't return nan when `sr == 0`. A sample Sharpe of exactly zero is a valid input, the correct output is a probability at or below 0.5.
- Please support 2D input like the other functions in this module. `mean`, `standard_deviation`, `skew` and `kurtosis` already work on 2D over axis 0, and the observation count becomes `np.sum(~np.isnan(returns), axis=0)`. Being able to score a whole (n_observations, n_trials) array in one call is useful on its own.
- Use `import scipy.stats as st` (the alias we use in `distribution/copula/_utils.py`) instead of `sct`.
- Not supporting `sample_weight` is fine, but mention it in the docstring since every other function here has it.
- ddof=1 for the std of the trial Sharpes is fine (it's the module default; the paper doesn't specify).

Docstrings:

- "Values close to 1 mean the winner clears the bar its own search sets by chance" says the opposite of what DSR means. It should be: the probability that the true Sharpe ratio is greater than zero after correcting for the number of trials, to compare against a confidence level (e.g. reject at 5% when DSR < 0.95).
- Please align the wording with the rest of the codebase: neutral tone, formulas in :math: notation, shorter comments. Same for the test comments.
- State that all quantities are per-period, not annualized (the paper's example uses SR = 2.5/sqrt(252) and trial variance 0.5/252; annualized trial Sharpes would give a wrong hurdle), and that `trial_sharpe_ratios` must be excess Sharpes when `risk_free_rate != 0`.
- The note about correlated trials making the estimate conservative is correct, keep it.

Tests: keep the skill-less search test. Please also add the worked example from the paper as a fixed-value test: `expected_max_sharpe_ratio(100, np.sqrt(0.5 / 252)) * np.sqrt(252)` should give about 2.53, and with SR = 2.5/sqrt(252), T = 1250, skew = -3, kurtosis = 10 the DSR is 0.8997. Add a test that `deflated_sharpe_ratio` equals `probabilistic_sharpe_ratio` with the expected max as benchmark, and one for the single-trial reduction to PSR. The zero-dispersion sweep can become a simple constant-series check once the guard is min == max.

Two last things: add the three functions to `docs/api.rst`, and GitHub shows a hidden/bidirectional Unicode warning on this branch, please check where that comes from.

ipezygj:
Thanks for the detailed review. All of it is in, pushed as one commit.

**Structure.** `probabilistic_sharpe_ratio` is its own measure with the signature you gave, and `deflated_sharpe_ratio` is now just PSR benchmarked against `expected_max_sharpe_ratio`. `expected_max_sharpe_ratio` keeps `(n_trials, trial_sharpe_ratio_std)` and gains `trial_sharpe_ratio_mean=0.0`, which it returns when `n_trials == 1` or the dispersion is not positive; `deflated_sharpe_ratio` always leaves it at the default, so the reduction to PSR against 0 holds. Your point about the correlated-trials case needing no new parameter is in the `expected_max_sharpe_ratio` notes: estimate an effective count, pass it as `n_trials`.

**Code.** Guard is `np.nanmin(returns, axis=0) == np.nanmax(returns, axis=0)`; the eps calibration and its comment are gone. `sr == 0` no longer returns nan. 2D over axis 0 is supported, with `n_observations = np.sum(~np.isnan(returns), axis=0)`. `import scipy.stats as st`. `sample_weight` is documented as unsupported. `ddof=1` unchanged.

**Docstrings.** The DSR description now says what it is: the probability that the true Sharpe ratio is greater than zero after correcting for the number of trials, compared against a confidence level (reject at 5% when DSR < 0.95). Formulas in `:math:`, per-period stated on both functions, excess trial Sharpe ratios stated for non-zero `risk_free_rate`, correlated-trials note kept.

Two things worth flagging.

**The 2.53.** `expected_max_sharpe_ratio(100, np.sqrt(0.5 / 252)) * np.sqrt(252)` gives **1.7894**, not 2.53. 2.5306 is the bracket alone, i.e. the value at unit dispersion:

```python
expected_max_sharpe_ratio(100, 1.0)                            # 2.5306
expected_max_sharpe_ratio(100, np.sqrt(0.5 / 252)) * np.sqrt(252)  # 1.7894 = 2.5306 * sqrt(0.5)
```

1.7894 is the paper's annualized hurdle for that example, and it is the number that produces your 0.8997 downstream. Both are now fixed-value tests.

**The 0.8997 test needed a constructed series.** Skew -3 and kurtosis 10 sit exactly on the boundary `kurtosis = skew**2 + 1`, so no distribution other than a two-point one has them: solving gives `p*q = 1/13`. The test therefore builds the series from two atoms (105 and 1145 of 1250), which lands within 0.004 of both moments and reproduces 0.8997 to four decimals through `probabilistic_sharpe_ratio`. Also added: DSR equals PSR against the expected max, the single-trial reduction, and 2D against column-wise 1D. The zero-dispersion sweep is now a plain constant-series check.

`docs/api.rst` lists the three functions. 260 tests pass in `test_measures.py`, ruff format and check are clean.

**On the hidden/bidirectional Unicode warning:** I cannot find it. The three files in this PR contain no bytes above U+007F at all, and a scan of every tracked file in the repo for bidi and invisible-format characters (Cf/Cc, RLO/LRO/PDF/isolates, ZWJ/ZWNJ/ZWSP/word-joiner, BOM) returns nothing outside the binary assets under `docs/_static` and `src/skfolio/datasets/data`. Could you point me at the file or the warning banner? If it is still showing after this push I would like to chase down what triggers it.

matteoettam09:
Thanks for landing PSR / DSR / `expected_max_sharpe_ratio` with this structure @ipezygj

We hit a related gap on the [Orion](https://www.orionfinance.ai/) side and had to implement a small scoring path ourselves because skfolio did not yet expose the non-probabilistic companions that sit under the same Bailey & López de Prado machinery. Sharing in case any of it belongs here next, happy to import from skfolio instead of keeping a fork of the math:

https://github.com/OrionFinanceAI/orion-finance-sdk-py/commit/d6e017d9dfec674243c1e5d610e313c8d5d149ae

**TLDR.** For ranking many short return series (not selection among backtests), we want a *score*, not a probability. We use:

1. **Distribution-adjusted Sharpe (DASR)** — the point-estimate haircut that reuses the same variance term already in your PSR denominator:

   \[
   \widehat{v}(\widehat{SR}) = 1 - \gamma_3\,\widehat{SR} + \frac{\gamma_4 - 1}{4}\,\widehat{SR}^2,
   \qquad
   \mathrm{DASR} = \frac{\widehat{SR}}{\sqrt{\widehat{v}(\widehat{SR})}}
   \]

   when \(\widehat{v} > 0\). Same skew / Pearson kurtosis convention as your `probabilistic_sharpe_ratio` (`variance_adj`). Refs: Sharpe (1994); Bailey & López de Prado (2012), *The Sharpe Ratio Efficient Frontier*.

2. **Lo effective sample size** — lag-1 autocorrelation haircut on \(T\):

   \[
   T_{\mathrm{eff}} = \frac{T}{1 + 2\hat\rho_1}
   \]

   (clipped so the factor stays positive). Ref: Lo (2002), *The Statistics of Sharpe Ratios*.

Those two pieces feel standard enough for `skfolio.measures`. A thin wrapper that returns DASR, and optionally accepting `n_observations=T_eff`, would let PSR and DSR stay correct under autocorrelation without inventing a new API shape.

We also linearly ramp the score from 0 to full DASR over a short track-record horizon (product default ~3 months of effective history before we fully trust the sample). That is conceptually related to PSR / MinTRL, may belong here as well. See how we use it here:

https://github.com/OrionFinanceAI/orion-finance-sdk-py/blob/d9d325393f0c2000eaae1b15fc463ec03f331f90/notebooks/intro_analytics.ipynb

Unrelated to the PR, but since we’re already in this thread. Your Ranger Sovereign Vault + Chimera work is very much the profile of people we want as managers / strategists at Orion! If you’re curious about running strategies through an ERC-7540 vault with onchain rebalancing, happy to chat offline (or drop a note on our side and I’ll follow up).

ipezygj:
Thanks for the detailed write-up - useful to see what the fork actually needs.

On the two asks:

**DASR.** Cheap to expose, since `v(SR)` is already computed inside
`probabilistic_sharpe_ratio`'s denominator. Worth being explicit about what it is,
though: `SR / sqrt(v(SR))` is the PSR z-statistic with `SR* = 0` and the `sqrt(T-1)`
factor removed. Dropping that factor is the whole point for your use case, but it means
DASR carries no sample size at all. If skfolio ships it as a measure, people will use
it to compare series of different lengths, which is exactly the comparison it can't
make. PSR and DSR carry T on purpose. So if it goes in, I'd want the docstring to say
that in one sentence rather than leave it implied.

**`n_observations=T_eff`.** No objection to the parameter. Passing T explicitly is
strictly better than the current implicit `len(returns)`, and it's a small change.

Computing `T_eff` inside skfolio is a different question and I'd keep it out of this
PR. `T / (1 + 2*rho1)` is unbounded: the denominator crosses zero at `rho1 = -0.5`, so
a library helper would need a policy for what happens on the other side of that, and
whatever it picks will be wrong for someone. A caller who has thought about their own
data can pass the T they want. A default can't.

Separately, `T / (1 + 2*rho1)` is `1 + 2*sum(rho_k)` truncated at lag 1. Under AR(1)
the full sum has a closed form, `(1 + rho) / (1 - rho)`, and at `rho = 0.3` that is
`0.54*T` against the truncated `0.63*T`. If you're applying this to series where the
autocorrelation persists past one lag, the truncated version understates the haircut.

Small thing: the PR isn't merged, it's still open. A review from someone who has had to
reimplement this would help it along.

I also read `stats/ranking.py`. There's a consequence for `w` on short track records
that I don't think you'd want found in production, and it's a measurement rather than a
comment, so I've written to you about it separately.

tschm:
[People interested in this topic should also have a look into jebel-quant/jquantstats. ](https://github.com/Jebel-Quant/jquantstats). I am not sure all of this functionality should sit in the very same package.

CarloNicolini:
I also add this interesting resource, recently updated (24 aug 2026) from Lopez De Prado and Emilio Porcu, so one can check correctness of implementation.

https://papers.ssrn.com/sol3/papers.cfm?abstract_id=7198158

tschm:
I would take all portfolio analytics out of skfolio. This could be separate repository and could be a dev dependency of skfolio. skfolio should offer portfolio optimization. You could even think about removing factor-models etc into their very own packages.

HugoDelatte:
@tschm happy to brainstorm on this! 

On my side, I think splitting into multiple packages and repos would add quite a lot of overhead without a clear benefit.

Factor models and analytics are really linked to the rest of the API. A factor model feeds into the optimization estimators, and we can tune its parameters together with the strategy. Analytics are also used for evaluation, cross-validation and hyperparameter tuning. This PR is a good example: assessing a Sharpe ratio after trying many parameter combinations is directly related to model selection. These are all parts of the same workflow.

Otherwise, for users going through that workflow, they would still need the different packages together. We would then have to manage compatibility between their versions, coordinate releases and spread changes across several repos. A change that we can currently implement and test in one PR could require coordinated changes in two or three packages. I’m not sure what we gain in return.

Also, analytics would still need to be a runtime dependency wherever we use them for scoring and evaluation. Making them a dev dependency would not cover those uses.

I’m very much in favour of keeping clear module boundaries and making each component usable independently. But we can do that within skfolio. Someone can use the analytics without running an optimization, just as they can use a factor model on its own.

For me, the value of skfolio is having these pieces work well together, from estimation to optimization to evaluation.

tschm:
yes, let's discuss soon. I think skfolio is getting stronger to more open it gets. Give the user the technical possibility to feed in a function for validation rather than integrating another library for it. So, the philosophy could be "we provide a standard Sharpe ratio but by design you can feed in alternative metrics".

HugoDelatte:
@tschm yes, I agree with that. Users should be able to plug in their own metrics and validation logic.

We already support this through our scikit-learn compatible API, for example by passing custom scoring functions for evaluation and hyperparameter tuning, including through `make_scorer`. So users can already choose an alternative metric or bring one from another library.

But I’m sure we can make it even more open and easier to extend, happy to discuss where we can improve.

tschm:
this sounds to me your Sharpe Ratio variations should be scoring functions in the scikit-learn ecosystem. There are users that want to use a deflated Sharpe Ratio but have no need for a portfolio optimizer. i doubt they would install skfolio with its heavy emphasis on portfolio construction.

PR #256, opened 2026-08-12
[BUG] ShrunkMu: use Hermitian eigenvalue solver to keep mu_ real

albystack:
#### Reference Issues/PRs
No issue exists; reproduced on `main` at commit c06db84.

`ShrunkMu` with the default James-Stein method returns a complex128
`mu_` (and complex `alpha_`, `beta_`) because `np.linalg.eigvals`
returns complex eigenvalues for real symmetric covariance matrices.
The complex dtype propagates into `EmpiricalPrior.return_distribution_.mu`
and breaks `EntropyPooling` views (e.g. `"AAPL >= prior(AAPL) * 1.2"`
becomes `"AAPL >= 0.0011375282128741+0.0000000000000000j"`, which the
equation parser cannot parse).

Since covariance matrices are symmetric, the fix switches to
`np.linalg.eigvalsh`, which returns real eigenvalues, is the canonical
solver for Hermitian matrices and is numerically more stable. Numerical
outputs are unchanged to machine precision (~1e-18); only the dtype is
corrected (complex128 -> float64).

No.

- Choice of `np.linalg.eigvalsh` vs `.real` on the output of `eigvals`.
- Backward-compatibility: numeric values are identical to the old real
  parts (verified against an independent James-Stein implementation).

Yes. `TestShrunkMu::test_james_stein_outputs_are_real` asserts
`mu_`, `alpha_`, `beta_` are real. It fails on `main` (complex dtype)
and passes with this PR. The pre-existing `test_shrinkage_mu` numeric
assertions still pass unchanged.

#### Any other comments?
This also fixes the currently-failing pushed test
`test_mean_views_prior_estimator[TNC]` and `[CLARABEL]` in
`tests/test_prior/test_entropy_pooling.py`.

#### PR checklist

### DIFF
--- a/src/skfolio/moments/expected_returns/_shrunk_mu.py
+++ b/src/skfolio/moments/expected_returns/_shrunk_mu.py
@@ -211,7 +211,10 @@ class ShrunkMu(BaseMu):
         # Calculate Estimators
         match self.method:
             case ShrunkMuMethods.JAMES_STEIN:
-                eigenvalues = np.linalg.eigvals(covariance)
+                # Covariance matrices are symmetric, so we use the Hermitian solver
+                # to get real eigenvalues. `np.linalg.eigvals` would return complex
+                # eigenvalues, making `mu_`, `alpha_` and `beta_` complex.
+                eigenvalues = np.linalg.eigvalsh(covariance)
                 self.beta_ = (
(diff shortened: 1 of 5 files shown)


HugoDelatte:
Good spot! I agree with the changes.
Thanks for the PR.
We have unrelated linting errors that I will fix before merging this PR.

Issue #150, opened 2025-06-08
[ENH] Add Geodesic shrinkage covariance estimators


HugoDelatte:
See https://github.com/skfolio/skfolio/discussions/3#discussioncomment-13148079

lcrmorin:
@HugoDelatte picking this up as this is a relatively easy update — opening a draft PR now.

PR #259, opened 2026-08-25
feat(moments): add GeodesicShrinkageCovariance estimator

lcrmorin:
Add an SPD-manifold geodesic interpolation as an alternative to linear (Euclidean) shrinkage between a covariance estimate and a target matrix (identity, diagonal, or a user-supplied SPD matrix). Implements the geodesic shrinkage family discussed in #150.


HugoDelatte:
Thanks for the PR @lcrmorin, 
I already had a quick look, and it’s very clean. I’ll do a proper full review over the coming weeks.

HugoDelatte:
The two CI failures are unrelated to this PR, and have been fixed on main. Merging main will solve them.

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+
+    Classical shrinkage estimators (e.g. :class:`~skfolio.moments.ShrunkCovariance`,
+    :class:`~skfolio.moments.LedoitWolf`) regularize the sample covariance matrix `S`
+    towards a well-conditioned target `T` using a linear (Euclidean) combination:
+
+    .. math:: (1 - \alpha) \cdot S + \alpha \cdot T
+
+    `GeodesicShrinkageCovariance` instead moves `S` towards `T` along the
+    affine-invariant geodesic of the manifold of Symmetric Positive Definite (SPD)
+    matrices [1]_ (see :func:`~skfolio.utils.stats.cov_geodesic_interpolation`):
```
HugoDelatte, line 37: Maybe slightly clearer:

```python
"""
`GeodesicShrinkageCovariance` instead moves `S` towards `T` along the geodesic
of the manifold of Symmetric Positive Definite (SPD) matrices under the
affine-invariant Riemannian metric (AIRM) [1]_ [2]_:
"""
```

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+    eigenvalues, :math:`\lambda_i^{1-\alpha} \cdot \mu^{\alpha}`, whereas linear
+    shrinkage is an *arithmetic* interpolation,
+    :math:`(1 - \alpha) \cdot \lambda_i + \alpha \cdot \mu`. In particular, for the
+    identity target the arithmetic interpolation lowers the condition number faster
+    than the geometric one as `shrinkage` increases from 0, since a small `shrinkage`
+    already lifts the smallest eigenvalues by an additive :math:`\alpha \mu` term.
+    Which interpolation is preferable is an empirical question left to the user;
+    both are made available so they can be compared, in keeping with skfolio's
+    approach of not imposing a default regularization on optimizers such as
+    :class:`~skfolio.optimization.SchurComplementary`.
```
HugoDelatte, line 64: I think we can make it a bit shorter, drop the ref to the GitHub discussion.
And drop the "lowers the condition number faster" sentence as it doesn't hold in general: with eigenvalues {10, 1, ..., 1} (twenty ones) and shrinkage=0.1, the geodesic gives cond 7.94 vs 8.77 for linear. 

```python
"""
At `shrinkage` :math:`\alpha = 0`, the estimate is the unmodified covariance from
`covariance_estimator`. At :math:`\alpha = 1`, it is exactly `target`. The
estimate remains SPD for every `shrinkage` in [0, 1] whenever `S` and `target`
are themselves positive definite.

The two schemes differ in how they treat the eigenvalues between the endpoints:
when `target` commutes with the sample covariance (e.g. the default
`SCALED_IDENTITY` target :math:`\mu \cdot I` with
:math:`\mu = \mathrm{trace}(S) / n`), the geodesic interpolation is a
*geometric* interpolation of the eigenvalues,
:math:`\lambda_i^{1-\alpha} \cdot \mu^{\alpha}`, whereas linear shrinkage is an
*arithmetic* interpolation, :math:`(1 - \alpha) \cdot \lambda_i + \alpha \cdot \mu`.
For the `SCALED_IDENTITY` target, the condition number of the geodesic estimate
is exactly :math:`\kappa(S)^{1-\alpha}`, and the trace is not preserved at
intermediate `shrinkage` values, unlike linear shrinkage.
"""
```

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+        :ref:`Covariance estimator <covariance_estimator>` used to compute the
+        starting covariance matrix `S` (the estimate returned when `shrinkage=0`).
+        The default (`None`) is to use :class:`~skfolio.moments.EmpiricalCovariance`.
+
+    shrinkage : float, default=0.1
+        Shrinkage intensity `alpha` in the geodesic interpolation. Must be between 0
+        (no shrinkage) and 1 (fully shrunk to `target`) inclusive. The default value
+        is `0.1`.
+
+    target : str or array-like of shape (n_assets, n_assets), default="identity"
```
HugoDelatte, line 78: I think we should use an AutoEnum here, following skfolio convention. 
With `target: GeodesicShrinkageTarget | ArrayLike`. Strings keep working since AutoEnum subclasses str. And maybe we should choose SCALED_IDENTITY over "identity" since the target is mu * I rather than I.
That Enum can be placed on top of the file and exposed in `__all__`.

```python
class GeodesicShrinkageTarget(AutoEnum):
    SCALED_IDENTITY = auto()
    DIAGONAL = auto()
```

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+        The shrinkage target `T`:
+
+            - "identity": :math:`\mu \cdot I`, with :math:`\mu = \mathrm{trace}(S) / n`.
+              This is the same default target used by scikit-learn's
+              `ShrunkCovariance`. Its condition number is exactly 1, so `shrinkage`
+              controls how far the estimate moves from `S` towards a perfectly
+              well-conditioned matrix.
+            - "diagonal": :math:`\mathrm{diag}(S)`, i.e. the correlations are shrunk
+              towards zero while each asset's variance is left unchanged.
```
HugoDelatte, line 87: The variances are only preserved at the endpoints. For example, at shrinkage=0.5 on the S&P 500 sample covariance they move by up to 16%.

```python
       - `DIAGONAL`: :math:`\mathrm{diag}(S)`. At `shrinkage=1` the
          variances equal those of `S` and all correlations are zero. At
          intermediate values the geodesic also changes the variances.
```

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+        has assets names that are all strings.
+
+    References
+    ----------
+    .. [1]  "Positive Definite Matrices".
+        Bhatia, R. (2007). Princeton University Press.
+
+    .. [2]  "Gracefully link HRP and min var", GitHub Discussion.
+        Cotton, P. & Delatte, H. (2024-2025).
+        https://github.com/skfolio/skfolio/discussions/3
```
HugoDelatte, line 133: I would drop the ref to the GitHub Discussion, and replace with the closest academic reference:


```
.. [2]  "Geodesically parameterized covariance estimation".
    Musolas, A., Smith, S.T. & Marzouk, Y. (2021).
    SIAM Journal on Matrix Analysis and Applications.
```

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+        -------
+        self : GeodesicShrinkageCovariance
+            Fitted estimator.
+        """
+        if not isinstance(self.shrinkage, numbers.Real) or not (
+            0.0 <= self.shrinkage <= 1.0
+        ):
+            raise ValueError(
+                f"shrinkage must be a float between 0 and 1, got {self.shrinkage}"
+            )
```
HugoDelatte, line 192: We have a helper for this in utils.tools and it also rejects booleans and NaN (shrinkage=True currently passes):

```python
_validate_unit_interval(self.shrinkage, "shrinkage")
```

Add it to the `skfolio.utils.tools` import and drop `numbers`.

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+
+        routed_params = skm.process_routing(self, "fit", **fit_params)
+
+        self.covariance_estimator_ = check_estimator(
+            self.covariance_estimator,
+            default=EmpiricalCovariance(),
+            check_type=BaseCovariance,
+        )
+        # noinspection PyArgumentList
```
HugoDelatte, line 201: the `# noinspection` comments are beeing removed from skfolio, so we can drop it.

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+        self.covariance_estimator_.fit(X, y, **routed_params.covariance_estimator.fit)
+
+        # we validate and convert to numpy after all models have been fitted to keep
+        # features names information.
+        X = skv.validate_data(self, X)
+        n_assets = X.shape[1]
+        start = np.asarray(self.covariance_estimator_.covariance_)
```
HugoDelatte, line 208: `nearest=True` on this estimator only applies after the interpolation, so a singular start with fewer observations than assets raises "start must be positive definite" before it can be repaired. The defaults are safe because the nested estimator repairs first (`EmpiricalCovariance` default to `nearest=True`). Maybe we should apply `cov_nearest` to start here when `self.nearest is True`. WDYT?

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+
+        target = self._target_matrix(start=start, n_assets=n_assets)
+
+        if self.shrinkage == 0.0:
+            covariance = start
+        else:
+            covariance = cov_geodesic_interpolation(
+                start=start, end=target, alpha=self.shrinkage
+            )
```
HugoDelatte, line 217: The default target commutes with S, so this path can skip the second eigendecomposition and all eight matmuls with a single eigh. It's about 3 times faster.

Also, the shrinkage == 0.0 shortcut duplicates the alpha == 0.0 shortcut inside the helper and can be dropped once validation is unified.

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+
+        self._set_covariance(covariance)
+        return self
+
+    def _target_matrix(self, start: npt.NDArray, n_assets: int) -> npt.NDArray:
```
HugoDelatte, line 222: To follow skfolio new typing conventions, import FloatArray from skfolio.typing instead of npt.NDArray for the argument and return type (you can then drop the numpy.typing import). 
Also n_assets can be derived from start.shape[0]. 
And maybe rename to _build_target_covariance.

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+                f"matrix, got string {self.target!r}"
+            )
+
+        target = np.asarray(self.target, dtype=float)
+        if target.shape != (n_assets, n_assets):
+            raise ValueError(
+                "target must be a square matrix of shape "
+                f"({n_assets}, {n_assets}), got shape {target.shape}"
+            )
+        assert_is_symmetric(target)
```
HugoDelatte, line 241: The helper returns `end` unchecked at `alpha=1.0` so an indefinite custom target can be returned without raising: GeodesicShrinkageCovariance(shrinkage=1.0, target=[[1, 2], [2, 1]], nearest=False).fit(X) gives a `covariance_` with min eigenvalue -1.0, while shrinkage=0.5 raises. I think we should validate PD here for array targets so every shrinkage value behaves the same.

src/skfolio/utils/stats.py
```diff
@@ -409,6 +409,85 @@ def corr_to_cov(corr: FloatArray, std: FloatArray):
     return cov
 
 
+def cov_geodesic_interpolation(
```
HugoDelatte, line 412: I'd move this into _geodesic_shrinkage_covariance.py as a private helper (e.g. _geodesic_interpolation) since it has a single consumer. That also means removing the :func: cross-reference in the class docstring (the formula is already there) and moving TestCovGeodesicInterpolation into the covariance test module. If SchurComplementary needs it later we can promote it back.

src/skfolio/utils/stats.py
```diff
@@ -409,6 +409,85 @@ def corr_to_cov(corr: FloatArray, std: FloatArray):
...
+            "`start` and `end` must have the same shape, got "
+            f"{start.shape} and {end.shape}"
+        )
+    if not 0.0 <= alpha <= 1.0:
+        raise ValueError(f"`alpha` must be between 0 and 1, got {alpha}")
+
+    if alpha == 0.0:
+        return start.copy()
+    if alpha == 1.0:
+        return end.copy()
```
HugoDelatte, line 472: These shortcuts return before the PD checks, so the Raises section is only accurate for alpha in (0, 1). Fine to keep for speed once the estimator validates the target.

tests/test_moment/test_covariance/test_covariance.py
```diff
@@ -2548,3 +2564,93 @@ def test_fit(self, X):
...
+        X = rng.standard_normal((100, 4)) * 0.01
+
+        model = GeodesicShrinkageCovariance(shrinkage=1.0, target="identity")
+        model.fit(X)
+
+        base = EmpiricalCovariance(nearest=False).fit(X)
+        mu = np.trace(base.covariance_) / base.covariance_.shape[0]
+        expected_target = mu * np.identity(base.covariance_.shape[0])
+        np.testing.assert_allclose(model.covariance_, expected_target, atol=1e-10)
+
```
HugoDelatte, line 2596: Could you add a test_metadata_routing like the ones in TestDenoiseCovariance and TestDetoneCovariance (wrapping ImpliedCovariance().set_fit_request(implied_vol=True))? It also covers the two get_metadata_routing lines codecov flags.

tests/test_moment/test_covariance/test_covariance.py
```diff
@@ -2548,3 +2564,93 @@ def test_fit(self, X):
...
+        if target == "identity":
+            mu = np.trace(start) / start.shape[0]
+            end = mu * np.identity(start.shape[0])
+        else:
+            end = np.diag(np.diag(start))
+        expected = cov_geodesic_interpolation(start=start, end=end, alpha=shrinkage)
+
+        np.testing.assert_allclose(model.covariance_, expected)
+
+    def test_custom_array_target(self):
```
HugoDelatte, line 2620: Maybe add a symmetric but indefinite target (e.g. [[1.0, 2.0], [2.0, 1.0]]) at both shrinkage=0.5 and shrinkage=1.0.

tests/test_moment/test_covariance/test_covariance.py
```diff
@@ -2548,3 +2564,93 @@ def test_fit(self, X):
...
+        np.testing.assert_allclose(model.covariance_, expected)
+
+    def test_wraps_other_covariance_estimator(self, X):
+        model = GeodesicShrinkageCovariance(covariance_estimator=LedoitWolf())
+        model.fit(X)
+        assert isinstance(model.covariance_estimator_, LedoitWolf)
+        assert model.covariance_.shape == (20, 20)
+
+    @pytest.mark.parametrize("shrinkage", [-0.1, 1.1, "a"])
+    def test_invalid_shrinkage_raises(self, shrinkage):
```
HugoDelatte, line 2643: You can also add `True` and `float("nan")`

tests/test_utils/test_stats.py
```diff
@@ -588,6 +589,80 @@ def test_invalid_std(self):
             corr_to_cov(corr, std)
 
 
+class TestCovGeodesicInterpolation:
```
HugoDelatte, line 592: Once the helper moves into the covariance module this class should follow, renamed accordingly. While at it, maybe add a non-PD end at an interior alpha (the "end must be positive definite" branch is the remaining uncovered line).

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
```diff
@@ -0,0 +1,242 @@
...
+        routed_params = skm.process_routing(self, "fit", **fit_params)
+
+        self.covariance_estimator_ = check_estimator(
+            self.covariance_estimator,
+            default=EmpiricalCovariance(),
+            check_type=BaseCovariance,
+        )
+        # noinspection PyArgumentList
+        self.covariance_estimator_.fit(X, y, **routed_params.covariance_estimator.fit)
+
```
HugoDelatte, line 203: I think we should propagate the nested estimator's `location_` so the inherited `score` and `mahalanobis` use the estimated mean rather than falling back to zero:

```python
if hasattr(self.covariance_estimator_, "location_"):
    self.location_ = self.covariance_estimator_.location_
```

Could you also add `location_` to the Attributes section of the docstring and a small test, e.g. `shrinkage=0` with `EmpiricalCovariance` asserting `location_` matches and `score(X_test)` equals the empirical one?

HugoDelatte (changes requested):
Thanks for the PR! It's very clean overall. 
A couple of inline comments needs to be addressed before we can merge.
The two CI failures are unrelated and fixed on main, so merging main will clear them.

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
lcrmorin: Addressed in [ad63e2f](https://github.com/skfolio/skfolio/commit/ad63e2ff510d01d938d3903a73b6ed4d9128ed8d). Added and exported `GeodesicShrinkageTarget` with `SCALED_IDENTITY` and `DIAGONAL`, including string-value support and API documentation. I clarified the AIRM description, corrected the condition-number/trace and diagonal-variance claims, and replaced the discussion citation with the academic reference. The interpolation helper and its tests now live in the covariance module; typing and naming follow the project conventions, the endpoint-validation limitation is documented, and the `noinspection` comments are removed.

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
lcrmorin: Addressed in [af52be2](https://github.com/skfolio/skfolio/commit/af52be22fcd9efd3fe24bc646ce644284ae6a72a). The fitted estimator now propagates `location_` and clears a stale mean when refitted with an estimator that does not expose one. Added tests that compare `score` and Mahalanobis distances with the wrapped estimator at zero shrinkage, plus the requested metadata-routing test using `ImpliedCovariance().set_fit_request(implied_vol=True)`.

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
lcrmorin: Addressed in [3b53b60](https://github.com/skfolio/skfolio/commit/3b53b604f1ff7b29e31a6bc0c49eee0979253314). Shrinkage now uses `_validate_unit_interval`, including rejection of booleans and NaN. Custom targets must be finite, symmetric, and positive definite at every shrinkage value, including both endpoints. With `nearest=True`, the starting covariance is repaired before interpolation. Added regression tests for indefinite and singular targets, both repair algorithms, and the non-PD end branch in the private helper; its tests were moved in the preceding commit.

src/skfolio/moments/covariance/_geodesic_shrinkage_covariance.py
lcrmorin: Addressed in [6242498](https://github.com/skfolio/skfolio/commit/624249816e0b49f81eb48a21e34978cfa3f3d6a6). The scaled-identity path now uses a single eigendecomposition and reconstructs the covariance from the interpolated eigenvalues. Both paths share the helper’s endpoint handling, so the duplicate zero-shrinkage shortcut is gone. The reference tests exercise enum and string targets, and an added regression test checks the exact condition-number formula across shrinkage values.

lcrmorin:
I've pushed 4 commits embarking your 18 comments. Each commit add relevant tests. Let me know if you think something is missing.

HugoDelatte:
Thanks for addressing the review comments, all changes look good.
We have some numerical instability issues on some OS, I'll take care of that investigation and final improvements.

HugoDelatte (approved):
Everything looks good now.

I've just added regression coverage for:
- Default `fit` on the seed-10 rank-deficient returns.
- Both built-in targets and repair methods on that fixture.
- Independent accuracy references for repaired covariance matrices.

PR #263, opened 2026-08-26
[BUG] Keep target NaNs out of prices_to_returns threshold

JohnDReynolds:
#### Reference Issues/PRs

No linked issue.

`prices_to_returns` documents `nan_threshold` as the allowed percentage of
missing asset prices. When target or factor prices were supplied through `y`,
the implementation counted missing values across the joined `X` and `y` frame
but compared that count with the number of assets in `X`.

Consequently, missing target values could remove observations whose asset
prices were complete.

This change computes the missingness threshold from the `X` columns only.
Joining, filling, inception handling, duplicate-label handling, and return
calculation remain unchanged.

A regression test covers complete asset prices with target-only NaNs and
independently calculates the expected asset returns.

No.

- Whether selecting the first `n_assets` columns correctly isolates `X` for
  every supported join mode.
- Whether the regression test clearly captures the documented asset-only
  `nan_threshold` behavior.
- Whether the change remains appropriately limited to threshold calculation
  without altering other missing-data behavior.

Yes. The new `test_returns_nan_threshold_ignores_y` regression test failed
before the fix because the returned asset frame had shape `(2, 1)` instead of
the independently expected `(3, 1)`.

Post-fix verification:

- Focused test file: 3 passed.
- Complete preprocessing subsystem: 194 passed.
- Ruff lint: passed.
- Ruff format check: all 457 files formatted.
- Full suite: 6,761 passed, with one documented local TNC numerical-tolerance
  exception and one sandbox permission failure involving the real
  `~/skfolio_data` cache. The dataset test passed when only that cache path was
  safely redirected to a temporary directory. The isolated TNC residual was
  `2.390084012331281e-07`, within the documented local `5e-7` exception.

#### Any other comments?

This changes no public signature, dependency, export, or documentation. The
existing docstring already describes the intended asset-only threshold
contract.

##### For all contributions

- [x] The PR title starts with [BUG].

##### For new estimators


HugoDelatte (approved):
Good catch!
The fix is correct and the changes are clean and minimal.
Thanks for the PR!

PR #264, opened 2026-08-27
[DOC] Add GraphicalLassoCV covariance example

### LINKED ISSUE #63: GraphicalLassoCV Example
#### GraphicalLassoCV Example

Is it possible to have an example where the precision matrix obtained with GraphicalLassoCV is employed?

albystack:
#### Reference Issues/PRs

Fixes #63.

This PR adds a `GraphicalLassoCV` tutorial for mean-risk portfolio optimization.

`GraphicalLassoCV` estimates both a regularized covariance matrix (`covariance_`) and a sparse precision matrix (`precision_`). The example demonstrates how the estimator is used in skfolio, visualizes the resulting partial correlations, and compares it with empirical covariance and Ledoit-Wolf shrinkage in long-only minimum-variance optimization.

The investigation also clarifies the distinction between statistical and computational sparsity:

* A zero off-diagonal precision entry represents conditional independence under the Gaussian graphical-model interpretation.
* Sparse precision does not generally imply sparse covariance.
* Cholesky factorization may introduce fill-in.
* skfolio currently defines portfolio risk from `covariance_` and represents variance using its covariance factor in the SOCP formulation.
* Although covariance and precision are inverses at the exact graphical-lasso optimum, the fitted API does not guarantee that the returned numerical matrices are exact inverses. Replacing `covariance_` with a matrix implied by `precision_` could therefore silently alter the portfolio optimization problem.

The example reports structural sparsity diagnostics only and intentionally makes no end-to-end performance claim.

No.

* Whether the example correctly explains the interpretation of `precision_`.
* Whether the distinction between statistical sparsity and SOCP computational sparsity is sufficiently clear.
* Whether retaining `covariance_` as the authoritative portfolio-risk representation is the appropriate conclusion for the current API.

No production behavior changes. The example was executed end-to-end and the focused `GraphicalLassoCV` and `EmpiricalPrior` tests were run.

#### Any other comments?

I also investigated whether the sparse precision matrix could be exploited directly in the conic formulation.

If an exact precision matrix satisfies `Θ = Σ⁻¹` and admits a factorization `Θ = C Cᵀ`, portfolio risk can be represented exactly by introducing an auxiliary variable `y` satisfying `C y = w`, so that `‖y‖₂² = wᵀΣw`.

Experiments indicate that this representation can substantially reduce the conic problem size when the model starts from a genuinely exact sparse precision matrix.

However, this is not appropriate as a hidden optimization path for the current `GraphicalLassoCV` wrapper because `precision_` is not an explicit exact-inverse contract for the authoritative `covariance_`.

A possible future enhancement would therefore be an explicit precision-authoritative estimator or exact precision-factor representation, rather than special-casing `GraphicalLassoCV`.

That design is intentionally outside the scope of this PR.

##### For all contributions

* [x] The PR title starts with either [ENH], [MNT], [DOC], or [BUG].

##### For new estimators

Not applicable.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -606,6 +606,7 @@ TUTORIAL_ORDER = {
         "plot_15_mip_cardinality_constraints.py",
         "plot_16_mip_threshold_constraints.py",
         "plot_17_failure_and_fallbacks.py",
+        "plot_18_graphical_lasso.py",
     ),
     "factor_models": (
         "plot_characteristics_factor_model.py",
--- a/examples/mean_risk/plot_18_graphical_lasso.py
+++ b/examples/mean_risk/plot_18_graphical_lasso.py
@@ -1,0 +1,190 @@
+r"""Graphical Lasso Covariance Estimation.
+==========================================
+
+This tutorial shows how to use
+:class:`~skfolio.moments.GraphicalLassoCV` as the covariance estimator of a
+:class:`~skfolio.optimization.MeanRisk` optimization.
+
+The empirical covariance matrix can be unstable when the number of assets is large
+relative to the number of observations. Graphical Lasso [1]_ regularizes the inverse
+covariance, also called the precision matrix, by solving an optimization problem of
+the form:
+
+.. math::
+
+    \underset{\Theta \succ 0}{\operatorname{minimize}}\;
+    \operatorname{tr}(S\Theta) - \log\det(\Theta)
+    + \alpha \lVert\Theta\rVert_1,
+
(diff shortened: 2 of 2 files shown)


HugoDelatte:
Good spot on your last commit about Codecov. Do you mind opening a dedicated PR for it? And I'll work on it with you to bring the coverage above 95% so we stop failing it.

albystack:
Done — I extracted the Codecov authentication change into dedicated PR #267 and removed it from this branch. PR #264 is back to the two documentation-only files.

PR #265, opened 2026-08-27
[BUG] Validate both fitness array dimensions

mameikagou:
#### Reference Issues/PRs

No existing issue.

`dominate()` requires two one-dimensional fitness arrays, but its chained comparison fails to reject several invalid shapes. For example, `dominate(np.array([[2]]), np.array([[1]]))` returns `True` instead of raising the intended `ValueError`.

Check each input's dimension separately, preserving the existing error message and behavior for valid arrays.

No.

The dimension guard and regression cases for either or both invalid inputs.

Nine parameterized cases cover scalar, two-dimensional, and three-dimensional inputs. The sorting tests and related portfolio, population, and pre-selection tests pass locally: 14 tests total. Ruff lint and format checks pass for both changed files.

#### Any other comments?

Validated on Python 3.12 with the project's development dependencies. The full test suite and cross-platform matrix were not run locally.

##### For all contributions
- [x] The PR title starts with [BUG].


HugoDelatte (approved):
Thanks for the PR!
The fix is correct and the regression tests cover the issue well. 
Good catch.

PR #266, opened 2026-08-29
[BUG] Reject invalid WalkForward window sizes

JohnDReynolds:
#### Reference Issues/PRs

None. The open-issue and pull-request review found no direct overlap with `WalkForward` window-size validation.

Reject non-integral or non-positive `test_size`, non-positive integer `train_size`, and negative `purged_size` consistently before split generation or split counting.

Previously, `WalkForward(test_size=0, train_size=2)` raised `ZeroDivisionError` from `get_n_splits`, while `split` could produce the same empty test fold indefinitely.

No. This change adds no runtime or development dependency.

Please review that `_validate_window_sizes`:

- Gives `split` and `get_n_splits` identical validation
- Accepts NumPy integral scalars while rejecting booleans
- Preserves constructor parameters stored on `self`
- Leaves valid calendar-offset schedules unchanged

Fractional `purged_size` typing and offset-mode purge placement remain separate follow-up work.

Yes. Twenty-four regression cases cover invalid sizes and NumPy integral values across both public methods.

#### Regression evidence

The regression tests failed before the fix because invalid sizes did not consistently raise from `split` and `get_n_splits`. All 24 cases pass after introducing the shared validation path.

The independent valid schedule for `train_size=2`, `test_size=2`, and `purged_size=1` remains:

- `([0, 1], [3, 4])`
- `([2, 3], [5, 6])`
- `([4, 5], [7, 8])`

The separately scoped fractional-purge and offset-purge behaviors remain unchanged.

#### Any other comments?

Verification completed with:

- Focused test file: 56 passed
- Model-selection subsystem: 302 passed
- Ruff, formatting, and changed-line Pyright passed
- Full suite: 6,786 passed with the documented local TNC tolerance exception
- TNC residual: `2.390084012331281e-07`, below the documented `5e-7` ceiling and reproduced in isolation

##### For all contributions

##### For new estimators

- [x] Not applicable—this PR does not add an estimator, so no API-reference entry is required.
- [x] Not applicable—this PR does not add an estimator requiring new docstring or gallery examples.


src/skfolio/model_selection/_walk_forward.py
```diff
@@ -256,6 +258,58 @@ def __init__(
...
+        if self.test_size < 1:
+            raise ValueError(f"`test_size` must be >= 1, got {self.test_size}.")
+
+        train_size = self.train_size
+        if isinstance(train_size, (bool, np.bool_)):
+            raise TypeError(
+                f"`train_size` must be an integer, got {type(train_size).__name__}."
+            )
+        # Integer windows must advance; calendar offsets retain their existing semantics.
+        if isinstance(train_size, numbers.Integral):
```
HugoDelatte: We can reuse `_is_integer_number` from `skfolio.utils.tools`. It also removes the `numbers` import and the asymmetry between the two guards (`test_size` checks bool only, `train_size` checks (bool, np.bool_) and only works because np.bool_ isn't registered with numbers.Integral).

src/skfolio/model_selection/_walk_forward.py
```diff
@@ -256,6 +258,58 @@ def __init__(
...
+            when `freq` is `None`.
+
+        ValueError
+            If `test_size` or an integer `train_size` is not positive, or if
+            `purged_size` is negative.
+        """
+        if isinstance(self.test_size, bool) or not isinstance(
+            self.test_size, numbers.Integral
+        ):
+            raise TypeError(
```
HugoDelatte: The codebase convention is `ValueError` for these (inheriting from scikit-learn convention). We should keep it for consistency. Same for Lines 295 and 304.

src/skfolio/model_selection/_walk_forward.py
```diff
@@ -256,6 +258,58 @@ def __init__(
...
+            if train_size < 1:
+                raise ValueError(f"`train_size` must be >= 1, got {train_size}.")
+        elif self.freq is None:
+            raise TypeError(
+                "`train_size` must be an integer when `freq` is None, got "
+                f"{type(train_size).__name__}."
+            )
+
+        if self.purged_size < 0:
+            raise ValueError(f"`purged_size` must be >= 0, got {self.purged_size}.")
```
HugoDelatte: `purged_size=1.5` or `True` still passes and fails later with an obscure numpy slicing error. We should also replace with `_validate_non_negative_integer(self.purged_size, "purged_size")`.

src/skfolio/model_selection/_walk_forward.py
```diff
@@ -256,6 +258,58 @@ def __init__(
...
+        # Integer windows must advance; calendar offsets retain their existing semantics.
+        if isinstance(train_size, numbers.Integral):
+            train_size = int(train_size)
+            if train_size < 1:
+                raise ValueError(f"`train_size` must be >= 1, got {train_size}.")
+        elif self.freq is None:
+            raise TypeError(
+                "`train_size` must be an integer when `freq` is None, got "
+                f"{type(train_size).__name__}."
+            )
```
HugoDelatte: With `freq` set, a `train_size` that is neither an integer nor an offset (e.g. "6M") passes validation and fails inside `_split_from_period_with_train_offset`. I think we should add an else branch requiring (pd.offsets.BaseOffset, dt.timedelta) and this also makes the bool special-case at line 294 unnecessary. Would require a new test also.

src/skfolio/model_selection/_walk_forward.py
```diff
@@ -256,6 +258,58 @@ def __init__(
...
+        if self.test_size < 1:
+            raise ValueError(f"`test_size` must be >= 1, got {self.test_size}.")
+
+        train_size = self.train_size
+        if isinstance(train_size, (bool, np.bool_)):
+            raise TypeError(
+                f"`train_size` must be an integer, got {type(train_size).__name__}."
+            )
+        # Integer windows must advance; calendar offsets retain their existing semantics.
```
HugoDelatte: The comment  is change-relative wording, let's drop it. Same for Lines 345 and 421.

tests/test_model_selection/test_walk_forward.py
```diff
@@ -441,6 +442,143 @@ def test_walk_forward_with_period_long(
...
+)
+@pytest.mark.parametrize("method_name", ["split", "get_n_splits"])
+def test_walk_forward_rejects_invalid_window_sizes(
+    test_size: object,
+    train_size: object,
+    purged_size: object,
+    error_type: type[BaseException],
+    match: str,
+    method_name: Literal["split", "get_n_splits"],
+) -> None:
```
HugoDelatte: I think we should drop the annotations and the cast(int, ...) calls.  They only silence the type checker on intentionally invalid values and don't match the style of the rest of this file. type[BaseException] would be type[Exception] anyway.
Same for Lines 4 , 550–553 and 570–573

src/skfolio/model_selection/_walk_forward.py
```diff
@@ -279,20 +333,26 @@ def split(
...
-            raise ValueError("test_size` must be an integer")
-
         if self.freq is None:
-            if not isinstance(self.train_size, int):
-                raise ValueError("When `freq` is None, `train_size` must be an integer")
             return _split_without_period(
                 n_samples=n_samples,
-                train_size=self.train_size,
-                test_size=self.test_size,
+                train_size=cast(int, train_size),
```
HugoDelatte: Let's drop the cast(int, ...) calls and the from typing import cast import, and pass train_size directly. skfolio doesn't run a static type checker and the helper already guarantees train_size is an int when freq is None. Same for line 427 and the int(first_valid) on line 464.

HugoDelatte (changes requested):
Thanks for the PR!
Nice catch and good centralization.
I've just requested some small changes, see inline comments.

JohnDReynolds:
Thank you for the detailed review. Your comments reinforced the importance of reusing existing utilities and following the codebase’s conventions. I also appreciate you catching the additional failure cases. I’ve addressed all seven points in the latest update.

HugoDelatte (approved):
Thanks for the changes, they look good.
I've just pushed a small commit to reuse the shared integer validators in WalkForward and report invalid values with !r so window-size errors match the tools helpers. And move _validate_window_sizes after the public class methods.

PR #269, opened 2026-08-31
[BUG] Apply WalkForward purging to offset training windows

JohnDReynolds:
#### Reference Issues/PRs

None.

This fixes `WalkForward` when `train_size` is a pandas offset or
`datetime.timedelta`. `purged_size` is now removed from the end of the training
window without shortening the test window.

Before this fix, with 120 daily observations, `train_size=pd.DateOffset(months=1)`,
`test_size=1`, `freq="MS"`, and `purged_size=1`, the first training fold incorrectly
remained `0..30` while the test fold was shortened from `31..58` to `31..57`.

After the fix, the training fold is correctly shortened to `0..29` and the test fold
remains `31..58`.

No.

- Confirm that offset-mode purging removes observations from training only.
- Confirm that test-window boundaries and fold counts remain unchanged.
- Confirm that the existing calendar-integer and observation-integer paths are unaffected.

Yes. The regression test covers:

- `pandas.DateOffset` and `datetime.timedelta` training windows.
- Both alignment directions.
- Rolling and expanding training windows.
- Complete and partial-final test policies.
- Preservation of test indices and fold counts.
- Independently calculated first-fold boundaries.

#### Regression evidence

- **Fail before:** The isolated regression failed because the purged training fold
  contained indices `0..30` instead of `0..29`; the implementation also shortened
  the test fold from `31..58` to `31..57`.
- **Independent expectation:** Remove only the final training observation, producing
  training indices `0..29` while preserving test indices `31..58`.
- **Pass after:** The isolated regression passed.
- The complete focused file passed all 72 tests.
- The complete model-selection subsystem passed all 318 tests.
- The final serial repository suite passed 6,958 tests. Its sole failure was the
  pre-existing local TNC numerical-tolerance case, independently reproduced with a
  residual of `2.390084012331281e-07`.
- Ruff lint, baseline-aware formatting, and `git diff --check` passed.
- Changed-line Pyright reported zero attributable diagnostics.

#### Any other comments?

The existing public documentation already states that `purged_size` excludes
observations from the end of training before testing, so no public documentation or
API changes were required.

##### For all contributions

##### For new estimators

Not applicable; this PR does not add an estimator.


HugoDelatte (approved):
Good catch! I agree with the fix.

PR #270, opened 2026-08-31
[BUG] Raise a clear error for empty cross-validation schedules

JohnDReynolds:
#### Reference Issues/PRs

None.

Detect when a cross-validator produces no splits and raise an actionable `ValueError` before shuffle or ordering inspection indexes the empty list.

Six observations with `WalkForward(train_size=5, test_size=2, reduce_test=False)` correctly produce zero complete folds, but `cross_val_predict` previously raised the internal error `IndexError: list index out of range`.

No.

Confirm that the empty-schedule guard runs immediately after split materialization, reports an actionable public error, and leaves every non-empty prediction path unchanged.

Yes. I added `test_cross_val_predict_rejects_empty_splits`, which independently verifies the zero-split premise and the new public error.

#### Regression evidence

- **Fail before:** Six observations with `WalkForward(train_size=5, test_size=2, reduce_test=False)` produce no complete splits, and `cross_val_predict` raises `IndexError: list index out of range`.
- **Independent expectation:** `cv.get_n_splits(X) == 0` and `list(cv.split(X))` is empty, so the public function should raise a stable `ValueError` explaining that the cross-validation strategy produced no splits.
- **Pass after:** The isolated regression, all 15 focused validation tests, and all 319 model-selection tests pass. Ruff lint, baseline-aware formatting, `git diff --check`, and changed-line Pyright also pass with no attributable diagnostics.
- **Unchanged control:** With `reduce_test=True`, the same observations still produce one partial fold and one portfolio.
- **Final suite:** The qualifying full run passed 6,959 of 6,960 tests. The sole failure was the pre-existing local TNC numerical-tolerance exception, independently reproduced at residual `2.390084012331281e-07`; no other test failed or errored.

#### Any other comments?

No public signature, export, dependency, or `WalkForward` scheduling behavior changes.

##### For all contributions

##### For new estimators

Not applicable; this PR does not add an estimator.


HugoDelatte (approved):
Thanks for the PR, looks good!

Issue #271, opened 2026-09-01
Code quality assessment at v1.0.4 — 7 tracked findings


tschm:
Tracking issue for a full code-quality assessment of the repository at `69c5234` (v1.0.4).

Each finding below is a sub-issue. This issue carries the shared context, the measurement
method, and the caveats — worth reading before picking up any child.

## How this was measured

Gates were resolved from the repo's **own committed configuration** only; no thresholds
were supplied from outside. Where a tool had no config in this repo (mypy, interrogate,
bandit) it was left unmeasured rather than run with invented settings.

| Gate | Resolved via | Verdict |
|---|---|---|
| Formatting / lint | `.pre-commit-config.yaml` (prek) | **FAIL** — config stale |
| Formatting / lint | `[tool.ruff]` as CI runs it | **PASS** — 0 errors, 460 files formatted |
| Type checking | no config found | unmeasured |
| Docstring coverage | ruff `D` rules, numpy convention | **PASS** — zero D1xx |
| Dependencies | `deptry` in the project env | **PASS** — 24 DEP002, no DEP001/003 |
| Tests | CI's own invocation | **FAIL** — 1 failed, 3 errors, 6947 passed |
| Coverage | `--cov=skfolio` vs `codecov.yml` | **PASS** — 96% vs 95% bar |
| Doc examples | doctest/fence parser | **FAIL** — 1 malformed, 83% never executed |

Environment: macOS arm64 (Darwin 24.6.0), Python 3.14.6, dependencies resolved latest via
`uv pip install -e ".[dev]"`. The CI matrix covers macOS only on Python 3.13, so
platform-sensitive results are flagged as such in the relevant child issues.

Two apparent failures were **discarded as measurement artifacts** rather than filed:
`deptry` run outside the project env reported 247 phantom "missing sklearn/cvxpy" errors
purely because the dependencies were not installed (24 real DEP002 once run correctly),
and the bundled README checker found nothing because it looks for `README.md` while this
project ships `README.rst`.

## Scorecard

| Subcategory | Score | Evidence |
|---|---|---|
| Docstring / API coverage | **10** | ruff `D` + numpy convention passes; only D205/D401 fire, both in the project's ignore list — zero D1xx across 460 files |
| Test coverage & depth | **9** | 96% (19259 statements, 747 missed) against the 95% project bar in `codecov.yml` |
| Semver discipline | **9** | semantic-release, `tag_format`, 53KB changelog, documented `mode = "update"` rationale |
[...]

Issue #277, opened 2026-09-01
test(prior): test_cvar_variance_views[TNC] fails deterministically on Python 3.14


tschm:
**Subcategory:** test pass rate

`tests/test_prior/test_entropy_pooling.py::test_cvar_variance_views[TNC]` fails
deterministically, including when run in isolation:

```
$ uv run --extra dev pytest "tests/test_prior/test_entropy_pooling.py::test_cvar_variance_views" -q
FAILED tests/test_prior/test_entropy_pooling.py::test_cvar_variance_views[TNC]
1 failed, 1 passed in 2.53s
```

The `CLARABEL` parametrization passes; only `TNC` fails.

**Raised at:** `src/skfolio/prior/_entropy_pooling.py:1112`

```
skfolio.exceptions.SolverError: Dual problem with Solver 'TNC' failed. This typically
occurs when the specified views conflict or are overly extreme. ... Solver error:
Linear search failed
```

**Environment:** macOS arm64 (Darwin 24.6.0), Python 3.14.6, dependencies resolved
latest via `uv pip install -e ".[dev]"`. Reproduced on every one of four full-suite runs.

The CI matrix covers macOS only on Python 3.13, so this combination may not be exercised
there — the failure could be a scipy-version-dependent regression in `TNC`'s line search
rather than a skfolio bug. Worth confirming which before deciding whether the fix is a
tolerance adjustment, a solver-parameter change, or an xfail with a version marker.

**done when…** the test passes on Python 3.14 with current scipy, or is explicitly
skipped/xfailed with a documented version condition.

tschm:
Part of #271 (code quality assessment at v1.0.4) — see that issue for the measurement method, environment and caveats.

PR #279, opened 2026-09-01
refactor(utils): break package-level import cycles

### LINKED ISSUE #276: refactor(utils): break three package-level cycles into the estimator layer
**Subcategory:** overall architecture — **7 → 9**

`skfolio.utils` is the bottom of the dependency stack by usage — **209 import
references, more than any other subpackage** — yet three of its modules reach upward
into estimator-layer packages, forming package-level cycles.

**The three cycles:**

| Downward edge | Return edge |
|---|---|
| `src/skfolio/utils/_factor_tools.py:13` → `skfolio.linear_model` | `src/skfolio/linear_model/_cross_sectional/_base.py:16` → `skfolio.utils.validation` |
| `src/skfolio/utils/_factor_tools.py:14` → `skfolio.preprocessing` | `src/skfolio/preprocessing/_transformer/_cross_sectional/_cs_standard_scaler.py:25` → `skfolio.utils.stats` |
| `src/skfolio/utils/validation.py:15` → `skfolio.containers` | `src/skfolio/containers/_asset_panel/_fields.py:18` → `skfolio.utils.tools` |

These resolve at import time today only because each cycle closes at *submodule*
granularity — `utils.stats` itself does not import `preprocessing`. That makes them
latent rather than harmless: any future import added to `utils/__init__.py` or to one of
the leaf modules can turn a working import into an `ImportError`, and the failure will
appear far from the change that caused it.

**Context — the rest of the import discipline is good.** Across 256 modules there are
only **two** runtime function-local imports (`containers/_asset_panel/_view.py:327`,
`model_selection/_validation.py:488`), both minimal, both at point of use, both genuine
cycle-breakers. Every other deferred import is properly under `if TYPE_CHECKING:`. This
finding is about the three package-level edges, not the general pattern.

**Suggested fix:** move `_factor_tools.py` out of `utils` into the estimator layer that
uses it, and move the `AssetPanel`/`AssetPanelView` validation in `utils/validation.py:295`
onto `containers` itself (or behind a Protocol), so `utils` depends on nothing above it.

**done when…** no module under `src/skfolio/utils/` imports `linear_model`,
`preprocessing` or `containers` at runtime.

tschm:
#### Reference Issues/PRs

Fixes #276. Part of the assessment tracked in #271.

`skfolio.utils` is the most-imported subpackage in the project — **209 import
references, more than any other** — yet three of its modules reached *upward* into
estimator-layer packages, forming three package-level cycles:

| Downward edge | Return edge |
|---|---|
| `utils/_factor_tools.py` → `linear_model` | `linear_model/_cross_sectional/_base.py` → `utils.validation` |
| `utils/_factor_tools.py` → `preprocessing` | `preprocessing/…/_cs_standard_scaler.py` → `utils.stats` |
| `utils/validation.py` → `containers` | `containers/_asset_panel/_fields.py` → `utils.tools` |

These resolved at import time only because each cycle closed at *submodule*
granularity — `utils.stats` itself does not import `preprocessing`. That made them
latent rather than harmless: any import later added to `utils/__init__.py` or to one
of the leaf modules could turn a working import into an `ImportError`, and the failure
would surface far from the change that caused it.

**Two different fixes, because the two situations are different.**

**Neutralization — relocate the code to the layer that owns it.** The function bodies
move unchanged.

- **`_cross_sectional_neutralize`** moves to `linear_model/_cross_sectional/_utils.py`
  as **`_cs_neutralize`**, matching the existing `_cs_*` helper naming. It is a
  weighted cross-sectional regression shared by two workflows, so it lives with the
  package that owns that operation, importing `BaseCSLinearModel` and
  `CSLinearRegression` directly from their defining sibling modules.

- **`_neutralize_scores`** moves to a module-level private function at the bottom of
  `alpha/_base.py`, its only production caller.

- **`_neutralize_exposures`** moves to the bottom of
  `prior/_characteristics_factor_model.py`, alongside the existing helpers. Its
  standardization, in-place updates and processing order are unchanged.

- **`utils/_factor_tools.py`** keeps the factor/family name-resolution helpers
  (`_factor_name_maps`, `_resolve_factor_name`, `_expand_factor_names`,
  `_resolve_factor_subset`), which now import only `numpy` and `skfolio.typing`.
  `_powered_market_cap` had no caller and no test, and is removed.

`preprocessing` stays independent of `linear_model`, so this adds **no new edge
between packages**.

**The validator — defer the import instead of moving the function.**
`validate_asset_panel` stays in `utils/validation.py`, next to
`validate_cross_sectional_data`, and keeps its public import path:

```python
from skfolio.utils.validation import validate_asset_panel
```

`AssetPanel` and `AssetPanelView` are imported *inside* the function, with a comment
saying why. Annotations are already postponed in that module, so the signature is
covered by a `TYPE_CHECKING` import and the runtime import happens only when the
validator is actually called. That removes the `utils → containers` edge without
moving anything users import.

#### Result

`skfolio.utils` imports only `_constants`, `exceptions` and `typing`, and the package
graph has no cycles at all:

```
typing / _constants / exceptions
        ↑
      utils
      ↑            ↑
 containers   linear_model
      ↑            ↑
     base       alpha / prior
```

Checked by building the package-level graph from module-level imports only (skipping
function-body and `TYPE_CHECKING` imports, which cannot close an import cycle):

```
main   → utils imports: _constants, containers, exceptions, linear_model,
                        preprocessing, typing
         6 cycles: utils <-> containers, utils <-> linear_model,
                   utils <-> preprocessing
branch → utils imports: _constants, exceptions, typing
         no package-level import cycles
```

#### Backward compatibility

None needed. `validate_asset_panel` never moves, so there is no deprecation, no shim,
and no caller churn — the earlier revision of this PR rewrote 46 import sites and added
a `FutureWarning` shim; both are gone. The public API is unchanged.

`validate_asset_panel` and `validate_cross_sectional_data` are now documented under a
new `skfolio.utils.validation` section in `docs/api.rst`. The module had no API page
before, which is why the function was not previously reachable from the docs.

No.

- **The deferred import in `validate_asset_panel`** — whether the comment explains the
  constraint well enough that someone does not later "tidy" it back to module scope.
- **`_cs_neutralize`'s new home**, and whether `linear_model/_cross_sectional/_utils.py`
  is where you would look for it.

No new tests are needed: the relocation is behaviour-preserving and the public import
path is unchanged. `tests/test_utils/test_validation.py` is byte-identical to `main`,
which is the point — the validator's tests did not have to move.

The tests for the relocated neutralization helpers moved with them:

| Tests | New home |
|---|---|
| score neutralization | `tests/test_alpha/test_base.py` |
| exposure neutralization | import updated to `skfolio.prior._characteristics_factor_model` |

Full suite on macOS arm64 / Python 3.14.6 / latest deps, on current `main`:
**6967 passed, 0 failed**. `ruff check` and `ruff format --check` are clean.

#### Any other comments?

The net diff is 9 files. The previous revision touched 58, almost all of it mechanical
import churn that this approach avoids entirely.


tschm:
@HugoDelatte it seems the Linux min test is somewhat fragile?

HugoDelatte:
@tschm indeed, it's part of the same test suit that you also flagged in https://github.com/skfolio/skfolio/issues/277
I'll work on them

HugoDelatte:
CI fixed in https://github.com/skfolio/skfolio/pull/290

HugoDelatte (changes requested):
Thanks for the PR, I agree with the cleanup direction. 

I’d like a few changes to the placement and backward compatibility before merging.

`validate_asset_panel` sets `asset_names_` / `n_assets_` on the estimator and checks them on `reset=False`, so it belongs with estimator support. Please move it to `skfolio.base`, next to `BaseAssetPanelTransformer`, together with `_check_fields_exist` and `_bad_observation`. That module already imports `AssetPanel`. Add the function to `skfolio.base.__all__` and update internal callers to use:

```python
from skfolio.base import validate_asset_panel
```

`validate_cross_sectional_data` should stay in `skfolio.utils.validation`, and the proposed `containers/_validation.py` can be removed.

The old validator import needs to keep working until 2.0: people use it when writing custom descriptors, and it is already explicitly exported through `__all__`. Please add a module-level `__getattr__` in `utils/validation.py` that resolves this name through a deferred import from `skfolio.base` and returns the actual function object. Emit a `FutureWarning` with `stacklevel=2` and this message:

> `skfolio.utils.validation.validate_asset_panel` is deprecated and will be removed in version 2.0. Import it from `skfolio.base` instead.

Keep the deprecated name in the old `__all__` until 2.0 so wildcard imports continue working too, consistent with the existing deprecation in `utils/sorting.py`. Unrecognized attributes should raise `AttributeError`. Please test that both explicit and wildcard legacy imports warn and resolve to the canonical function, while the new import does not warn. Please also include the API documentation entry under `skfolio.base`.

For neutralization, I’d prefer to separate the shared regression operation from the workflows that use it:

- Move `_cross_sectional_neutralize` to `linear_model/_cross_sectional/_utils.py`, renamed `_cs_neutralize`, consistent with the existing `_cs_*` helper naming. Import `BaseCSLinearModel` and `CSLinearRegression` directly from their defining sibling modules.
- Move `_neutralize_scores` to a module-level private function at the bottom of `alpha/_base.py`, its only production caller.
- Move `_neutralize_exposures` to the bottom of `prior/_characteristics_factor_model.py`, alongside the existing helpers. Preserve its standardization, in-place updates, and processing order.
- Remove the proposed `preprocessing/_neutralize.py`. This keeps `preprocessing` independent of `linear_model`.

The intent is to keep single-consumer helpers with their callers and shared helpers with the package that owns their operation. `utils/_factor_tools.py` can retain the factor/family name-resolution helpers without importing estimator packages. `_powered_market_cap` has no caller or test, so please remove it.

Please move the tests accordingly: validator behavior tests alongside tests for `skfolio.base`, compatibility tests under `test_utils`, score-neutralization tests under `test_alpha`, and the exposure-neutralization test import to `skfolio.prior._characteristics_factor_model`.

Finally you can remove the note about a `prior ↔ factor_exposure` cycle. `factor_exposure/_base.py` imports `skfolio.base`; the `skfolio.prior` references in that package are docstring examples, so there is no corresponding runtime import edge.

HugoDelatte (changes requested):
All the neutralization changes look good now.

However, sorry for sending you in the wrong direction with my earlier suggestion to move `validate_asset_panel` to `skfolio.base`.

After deeper reflexion I think we should favour UX and API consistency: from a user’s perspective, I think `skfolio.utils.validation` is the most natural location because it describes the operation, keeps the function alongside `validate_cross_sectional_data`, and follows the scikit-learn convention.

So I think we should:

- kkeep the existing public import:

```python
from skfolio.utils.validation import validate_asset_panel
```

For the validator, please:

- Keep `validate_asset_panel`, `_check_fields_exist`, and `_bad_observation` in `utils/validation.py`.
- Revert the validator-related caller imports to their original paths.
- Remove the `skfolio.base` export and the deprecation shim. Keep the existing `utils.validation.__all__`.
- Keep the validator behavior tests under `test_utils`; the deprecation tests are no longer needed.
- Document the function under `skfolio.utils.validation` instead of `skfolio.base`.

To address the eager dependency on `containers` that you spotted, let's move the runtime import of `AssetPanel` and `AssetPanelView` inside `validate_asset_panel`, with a short comment explaining why. Keep their annotation imports under `TYPE_CHECKING`, using the module’s existing postponed annotations.

Apologies for the extra round of changes.

tschm:
No apology needed — the second placement is the better one, and it makes the PR considerably smaller.

All done as described:

- `validate_asset_panel`, `_check_fields_exist` and `_bad_observation` stay in `utils/validation.py`. `utils.validation.__all__` is unchanged.
- The `skfolio.base` export and the deprecation shim are gone, and all 46 caller imports are back to their original paths.
- Validator tests are back under `test_utils` — `tests/test_utils/test_validation.py` is now byte-identical to `main`, and `tests/test_base.py` is dropped along with the deprecation tests.
- `AssetPanel` / `AssetPanelView` are imported inside `validate_asset_panel` with a comment explaining the constraint; the annotations stay under `TYPE_CHECKING` on the module's existing postponed annotations.
- Documented under a new `skfolio.utils.validation` section in `docs/api.rst`. That module had no API page at all before, which is why the function wasn't reachable from the docs — `validate_cross_sectional_data` is listed too.

To check the cycles are actually gone rather than just relocated, I built the package-level graph from module-level imports only, skipping function-body and `TYPE_CHECKING` imports since neither runs at import time:

```
main    utils imports: _constants, containers, exceptions, linear_model,
                       preprocessing, typing
        6 cycles: utils <-> containers, utils <-> linear_model,
                  utils <-> preprocessing

branch  utils imports: _constants, exceptions, typing
        no package-level import cycles
```

Full suite 6967 passed, ruff clean.

Net diff is **9 files**, down from 58 — dropping the relocation removes the import churn you flagged in the first review, so there is no longer anything mechanical to skim past.

tschm:
Added `tests/test_import_cycles.py` so the layering is enforced rather than just documented — the fix rests on one import staying inside `validate_asset_panel`, and nothing stopped a later reader from tidying it back to module scope.

Two static checks:

- **`test_no_package_import_cycles`** builds the subpackage graph from imports that actually run at import time (function bodies and `TYPE_CHECKING` blocks skipped, class bodies kept) and fails on any strongly connected component.
- **`test_utils_does_not_import_estimator_layer`** is the sharper case: `utils` is imported by everything, so it may import nothing above it, cycle or not.

I checked the guard actually bites by reverting the deferred import — both fail, and the message points straight at the two lines rather than just saying "cycle detected":

```
Failed: skfolio subpackages import each other at import time.

cycle between: containers, utils
  containers -> utils
      skfolio/containers/_asset_panel/_fields.py:18 imports skfolio.utils.tools
  utils -> containers
      skfolio/utils/validation.py:17 imports skfolio.containers

Move the offending import into the function that needs it, or relocate the code
to the layer that owns it.
```

One thing worth flagging, since it looks like an omission: there is no `sys.modules` runtime assertion. I wrote one first and it "failed" on `measures`, `population` and `portfolio` — because importing any submodule executes `skfolio/__init__.py`, which imports the whole package, so at runtime every subpackage is loaded whichever one you asked for. The check has to be static to mean anything. That reasoning is in the module docstring so nobody re-adds it.

Full suite 6969 passed; the two checks add 0.13s. Happy to drop them if you would rather not have an architecture test in the suite.

HugoDelatte (approved):
Looks good now, thx!

Issue #272, opened 2026-09-01
fix(datasets): write the download cache atomically


tschm:
**Subcategory:** test pass rate — **5 → 10**

`download_dataset` writes the download straight to its final cache path, with an
`os.path.exists` check as the only guard. The write is therefore neither atomic nor
concurrency-safe.

**File:** `src/skfolio/datasets/_base.py:153-159`

```python
if os.path.exists(filepath):
    return load_gzip_compressed_csv_data(filepath)
...
ur.urlretrieve(url, filepath)
```

**Two failure modes, both observed:**

1. The test suite runs with `-n=4 --dist=worksteal` (`pyproject.toml` `[tool.pytest.ini_options]`).
   One worker calls `urlretrieve` while another sees `os.path.exists() == True` and reads the
   partially written file:
   `EOFError: Compressed file ended before the end-of-stream marker was reached`
   (raised from `tests/conftest.py:57` → `load_sp500_implied_vol_dataset()`).
2. **For end users:** any interrupted download leaves a truncated `.csv.gz` cached
   permanently. Every subsequent call fails identically until the user manually deletes
   `~/skfolio_data`. There is no validation and no recovery path.

**Evidence:** across three consecutive full runs the error count fell 28 → 3 → 1 as the
cache warmed — the signature of a partial-write race rather than a deterministic bug.

**Suggested fix:** `urlretrieve` to a temporary file in the same directory, then
`os.replace()` it into position (atomic on POSIX and Windows). Optionally validate the
gzip stream before promoting it.

**done when…** a full `pytest -n=4` run against an empty `~/skfolio_data` completes with
zero `EOFError` setup errors, and an interrupted download leaves no file in the cache.

tschm:
Part of #271 (code quality assessment at v1.0.4) — see that issue for the measurement method, environment and caveats.

PR #280, opened 2026-09-01
chore(lint): make pre-commit the single lint gate

### LINKED ISSUE #274: chore(lint): .pre-commit-config.yaml pins ruff v0.8.4 while CI resolves 0.16.5
**Subcategory:** linting & style — **6 → 10**; toolchain reproducibility — **5 → 8**

The repo enforces two different formatting standards, and only one of them is wired to
anything.

**Files:** `.pre-commit-config.yaml`, `.github/workflows/ci.yml`

- `.pre-commit-config.yaml` pins `astral-sh/ruff-pre-commit` at **`rev: v0.8.4`**.
- CI installs `ruff` unpinned from the `dev` extra and resolves whatever is current
  (**0.16.5** at time of writing).

Running the committed config against the committed tree rewrites **62 files**:

```
ruff-lint ....... Failed   Found 191 errors (10 fixed, 181 remaining)
ruff-format ..... Failed   26 files reformatted, 484 files left unchanged
end-of-file-fixer Failed   36 further files modified
```

Most of the 181 unfixable errors are `UP038`, a rule later ruff versions removed
entirely. Under the version CI actually uses, the tree is clean:

```
$ uvx ruff check   # 0.16.5
All checks passed!
$ uvx ruff format --check
460 files already formatted
```

**No CI job invokes pre-commit**, and `CONTRIBUTING.md` never mentions it — it documents
`uv run ruff check --fix` / `uv run ruff format` instead. So the file is orphaned: it
affects only contributors who install the hooks on their own initiative, and for them it
blocks commits over rules the project does not enforce.

**Suggested fix:** bump `rev` to the ruff release CI resolves and add a CI job that runs
the hooks (making the config a real gate), or delete `.pre-commit-config.yaml`.

**done when…** `prek run --all-files` on a clean checkout passes with no file
modifications, and a CI job invokes it — or the file is removed.

tschm:
#### Reference Issues/PRs

Fixes #274. Part of the assessment tracked in #271.

`.pre-commit-config.yaml` pinned `ruff-pre-commit` at **v0.8.4** while CI installs
`ruff` unpinned from the `dev` extra and resolves **v0.16.5**. The two enforced
different rules, and nothing noticed: no CI job invoked the hooks, and
`CONTRIBUTING.md` never mentioned them. Running the committed config against the
committed tree rewrote **62 files** — 191 lint errors (181 unfixable, mostly `UP038`,
a rule later ruff removed outright) and 26 files reformatted.

**Bumping the rev was not sufficient**, which is the interesting part. `[tool.ruff]
include` scopes CI to `src/` and `tests/`, but **ruff ignores `include` for paths
passed explicitly on the command line** — and that is exactly how pre-commit invokes
it. So even at v0.16.5 the hooks reported 175 errors, every one of them in
`examples/`:

| rule | count | what it actually is |
|---|---|---|
| `B018` useless-expression | 66 | the bare trailing expression that renders a sphinx-gallery cell's output |
| `D400` missing-trailing-period | 47 | gallery prose docstrings |
| `E402` import-not-at-top | 26 | deliberate `# %%` cell structure |
| `RUF002`/`RUF003` ambiguous unicode | 19 | maths symbols in comments |

These are gallery idioms, not defects, and the project already keeps that tree out of
scope. So the hooks and CI differed in **two** dimensions — ruff version *and* file
scope — and only fixing both makes them agree.

Changes:

- Bump `ruff-pre-commit` to **v0.16.5**, matching what CI resolves.
- Scope both ruff hooks with `files: ^(src|tests)/.*\.py$` so they check the same tree
  `include` does.
- **Run the hooks in the `lint` job, replacing the direct `ruff` invocation.** Keeping
  both would reintroduce the drift this PR removes: the hook pins a release while the
  `dev` extra floats, so the next ruff release desynchronises them again. One gate,
  one pinned version.
- Cache hook environments, keyed on the config hash so the cache invalidates exactly
  when a hook is bumped.
- **Add the `pre-commit` ecosystem to dependabot**, so the revs cannot go stale again.
  This is the part that stops #274 recurring rather than just fixing today's instance.
- Document `pre-commit install` in `CONTRIBUTING.md`, now that the hooks gate the build.

No. `pre-commit` is already in the `dev` extra; it simply was not being run.

- **Replacing the direct ruff call rather than adding pre-commit alongside it.** This
  is the main judgement call, and it costs you the `--output-format=github` inline PR
  annotations. I chose it because running ruff twice at two versions is the precise
  failure this issue reports. If the annotations matter more, the alternative is to
  keep the direct call and pin `ruff==0.16.5` in the `dev` extra — but then two places
  need bumping in lockstep, and dependabot will only move one of them. Happy to switch
  if you prefer that trade.
- **Whether `examples/` should be linted at all.** I preserved the existing decision
  (excluded) rather than overturning it. A defensible alternative is to lint the
  gallery with `per-file-ignores` for `B018`/`E402`/`D400`, which would catch real
  problems in example code that nothing currently checks. That is a bigger change and
  felt like yours to make.
- **`pre-commit-hooks` is still at v5.0.0** while v6.0.0 is out. I left it alone —
  a major bump is unrelated to this issue and carries its own behaviour risk.
  Dependabot will now propose it separately.

No — there is no source change to test. Verification was:

- `prek run --all-files` on a fresh checkout: **all four hooks pass, zero files
  modified** (previously 62 modified).
- `actionlint` clean on the modified workflow.
- Full test suite run; see below.

#### Any other comments?

**This branch touches no `src/` or `tests/` file.** The diff is `.pre-commit-config.yaml`,
`.github/` (workflow + dependabot), `CONTRIBUTING.md`, and 31 files whose only change is
end-of-file-fixer output — 7 gain a missing trailing newline (`codecov.yml`,
`.github/CODEOWNERS`, `docs/.gitignore` and others genuinely lacked one), the rest lose
a redundant trailing blank line.

The local test run reported `1 failed, 6929 passed, 29 errors`, and **none of it comes
from this branch**. The single failure is `test_cvar_variance_views[TNC]` (#277). All 28
errors are `EOFError: Compressed file ended before the end-of-stream marker`, plus one
`EmptyDataError: No columns to parse from file` — the dataset-cache corruption in #272.
`~/skfolio_data/` was visibly partial during the run, with two datasets missing and three
re-downloaded mid-suite. That bug has now produced a different error count on five of six
local runs (28 → 3 → 2 → 1 → 0 → 29), which is a fair argument for prioritising #272:
it makes every other result on this repo harder to read.


HugoDelatte (dismissed):
Thanks for the PR!

I’ve simplified the setup while keeping CI on the latest Ruff release:

- CI runs Ruff directly and local pre-commit hooks use Ruff from the development environment.
- Development and documentation tools use dependency groups. `uv.lock` stays local and ignored.
- Minimum/latest dependency tests are preserved, with installed versions logged for debugging.
- Contributor instructions are shorter and explain hook fixes and refreshing Ruff.
- Docker installs its existing notebook tools through an optional dependency group.

tschm:
@HugoDelatte It’s great that you’re testing the minimal versions. I wonder if we might be going a little too far back here, though. Personally, I’d lean towards Clarabel > 1.0 and NumPy > 2.0, and I’m not sure we need to support Python 3.10 at this stage. But of course, happy to leave that call to **you**

tschm:
it's actually a good idea to commit also uv.lock! That's the best way to make sure you and also others use the same virtual environment

HugoDelatte:
@tschm yep I agree I've bumped some of the min versions.
Concerning python 3.10, we will keep it at least until October to match CPython’s lifecycle.

HugoDelatte:
@tschm concerning uv.lock, 

There is a long-running debate around committing lockfiles for libraries ([uv discussion](https://github.com/astral-sh/uv/issues/10730), [Rust discussion](https://blog.rust-lang.org/2023/08/29/committing-lockfiles/)).

For applications I totally agree with you, but for libraries I think the trade-off is less clear. 

The main value of uv.lock here would be reproducibility for contributors. 
My concern is whether we actually want the default development environment pinned to a specific dependency resolution. As a library, we benefit from contributors and CI continuously exercising the current dependency resolution.

HugoDelatte (approved):

PR #281, opened 2026-09-01
chore: remove the broken, unreferenced Makefile

### LINKED ISSUE #273: chore(build): Makefile uv targets fail — uv run omits the dev extra
**Subcategory:** contributor onboarding — **6 → 9**

Both `uv` targets in the `Makefile` fail on a clean checkout.

**File:** `Makefile`

```
uv-format-and-lint:
	uv run ruff check --fix
	uv run ruff format

uv-test:
	uv run pytest
```

`uv run` installs only the `[project] dependencies`. `pytest` and `ruff` are declared in
the `[project.optional-dependencies] dev` extra, so neither is present:

```
$ make uv-test
uv run pytest
error: Failed to spawn: `pytest`
  Caused by: No such file or directory (os error 2)
make: *** [uv-test] Error 2
```

The `pip-*` targets assume an already-activated environment with dev extras installed.

Compounding it: **nothing references the Makefile.** `.github/workflows/ci.yml` runs
`uv pip install -e ".[dev]"` then `uv run pytest -q tests` directly, and
`CONTRIBUTING.md` documents the same explicit commands. So the Makefile is the only
entry point that does not work, and it is the one a newcomer is most likely to try first.

**Suggested fix:** either add `--extra dev` to the `uv run` invocations, or delete the
Makefile and let `CONTRIBUTING.md` remain the single documented entry point.

**done when…** `make uv-test` and `make uv-format-and-lint` succeed on a clean checkout
with no prior environment setup — or the Makefile is removed.

tschm:
#### Reference Issues/PRs

Fixes #273. Part of the assessment tracked in #271.

Both `uv` targets in the root `Makefile` failed from a clean checkout:

```
$ make uv-test
uv run pytest
Creating virtual environment at: .venv
Installed 19 packages in 205ms
error: Failed to spawn: `pytest`
make: *** [uv-test] Error 2
```

`pytest` and `ruff` live in the `[project.optional-dependencies] dev` extra, so the
environment `uv run` builds contains neither. The `pip-*` targets have the
mirror-image problem: they invoke `ruff` and `pytest` bare, silently assuming an
already-activated environment with dev extras installed.

**This PR removes the file rather than repairing it.** The evidence for that:

- **Nothing referenced it.** Not `ci.yml`, not `CONTRIBUTING.md`, not the docs, not the
  `Dockerfile`. `grep` across the repo for `make <target>` returns nothing.
- **It arrived by accident.** `git log -- Makefile` shows a single commit, fd38213 —
  *"fix: Adds support for one-factor models using EmpiricalCovariance (#188)"* — where
  it appears alongside exactly one source file. It was never a considered
  contributor-facing interface, and it has been broken since the day it landed.
- **`CONTRIBUTING.md` already covers everything it wrapped, and covers it correctly.**
  Its setup step is `uv venv` followed by `uv pip install --editable ".[dev]"`, and I
  verified the documented flow end to end: after those two commands, `uv run pytest`
  works. The Makefile failed precisely *because* it skipped that setup step, not
  because the documented workflow is wrong. So there is nothing to migrate.
- **Repairing it would create a third lint entry point**, competing with
  `.pre-commit-config.yaml` and CI — the same class of divergence as #274, where a
  stale hook pin and an unpinned CI ruff quietly enforced different rules.

A broken front door that nothing points at is worse than no front door.

No.

- **Whether you want a working `make` at all.** Some contributors reach for `make test`
  first, and that habit is exactly why a broken target is a trap. If you would rather
  keep the affordance, the alternative is a thin wrapper delegating to the canonical
  commands (`uv run --extra dev pytest`, `uv run pre-commit run --all-files`) plus a
  `help` target — happy to send that instead. I went with removal because nothing
  currently points at the file and `CONTRIBUTING.md` is already complete.

No — this deletes an unreferenced file. Verified that no test or packaging input
references it: `grep -rn "Makefile" tests/ src/` returns nothing, and `MANIFEST.in`
and `pyproject.toml` do not mention it.

#### Any other comments?

`docs/Makefile` and `docs/make.bat` are Sphinx's own scaffolding and are untouched. The
`.gitattributes` line `Makefile text eol=lf` is deliberately **kept**: gitattributes
patterns without a slash match at any depth, so that rule still applies to
`docs/Makefile`.

**Unrelated bug spotted while checking what else the Makefile might affect:** the
`Dockerfile` cannot build. It does

```dockerfile
RUN --mount=type=bind,source=uv.lock,target=uv.lock \
    uv sync --frozen --no-install-project --no-dev
```

but `.gitignore:93` contains `*.lock`, so `uv.lock` is never committed and the bind
mount has no source. Not touched here — say the word and I'll open a separate issue.


tschm:
@HugoDelatte this can be merged. It's just removing a Makefile that is incorrect and not used by the ci/cd process

HugoDelatte (approved):
Good spot! I agree it can be removed

PR #282, opened 2026-09-01
fix(datasets): write the dataset cache atomically

### LINKED ISSUE #272: fix(datasets): write the download cache atomically

tschm:
#### Reference Issues/PRs

Fixes #272. Part of the assessment tracked in #271.

`download_dataset` checked `os.path.exists(filepath)` and then handed that same path
straight to `ur.urlretrieve`, so the cache entry was written **in place**:

```python
if os.path.exists(filepath):
    return load_gzip_compressed_csv_data(filepath)
...
ur.urlretrieve(url, filepath)
```

Two consequences, both observed:

1. **Concurrent readers see a partial file.** The suite runs with `-n=4
   --dist=worksteal`, so one worker reads while another is mid-write:
   `EOFError: Compressed file ended before the end-of-stream marker was reached`.
2. **For end users, an interrupted download poisons the cache permanently.** The
   truncated `.csv.gz` stays, every later call fails identically, and there is no
   validation and no recovery — the only remedy is deleting `~/skfolio_data` by hand.

**The fix.** Download to a temporary file in the same directory and promote it with
`os.replace`, atomic on POSIX and Windows when both paths share a filesystem. The
cache entry now either does not exist or is complete. Cleanup is keyed on
`BaseException` so `KeyboardInterrupt` — the very way these truncated entries arise —
leaves nothing behind. The promoted file is chmod'd to `0644`, because `mkstemp`
creates `0600` and the cache can be shared between users via `SKFOLIO_DATA`.

**Plus recovery for caches that are already poisoned.** Atomic writes do nothing for
installations that already hold a truncated file, so an entry that fails to decompress
is now discarded and re-fetched. The guard is deliberately narrow — `EOFError`,
`gzip.BadGzipFile`, pandas parser errors — so permission and disk failures still
propagate rather than silently triggering a re-download. With
`download_if_missing=False` the error now names the unreadable file instead of
surfacing a bare gzip failure.

#### A second bug found while verifying this one

Three tests in `tests/test_dataset/test_dataset.py` operated on the **real**
`~/skfolio_data`: `TestGetDataHome::test_create_directory` wiped it with
`shutil.rmtree`, `TestClearDataHome::test_delete_content_default_path` called
`clear_data_home()` on it twice, and `test_delete_content_valid_path` created a
*relative* `valid/path` in the working directory.

Under `--dist=worksteal` those raced with every test that reads a dataset. Isolating
it makes the difference stark:

```
pytest tests/test_dataset/ -n0     ->  30 passed
pytest tests/test_dataset/         ->   1 failed   (FileNotFoundError)
```

This is what kept *triggering* the corruption above: a worker deleted the cache
mid-run, the remaining workers all re-downloaded at once, and the non-atomic write
did the rest. The two bugs compounded, which is why the error count swung so wildly
between runs. These tests now use `tmp_path` and a `SKFOLIO_DATA` fixture.

I have kept this in the same PR because it is the other half of the same failure and
is ~20 lines. Happy to split it out if you would rather review them separately.

No. `contextlib` and `tempfile` are stdlib.

- **The recovery path is a behaviour change**, not just a robustness fix: a corrupt
  cache entry is now silently deleted and re-downloaded rather than raising. I think
  that is right for a cache, but it is the judgement call in this PR.
- **`chmod 0o644` ignores the user's umask.** Preserving umask exactly would need
  `os.umask(0)` and a restore, which is not thread-safe. `0644` matches what a plain
  `urlretrieve` produces under a typical umask; flag it if you would rather keep
  `mkstemp`'s `0600`.
- **Narrowness of `_CORRUPT_CACHE_ERRORS`.** I chose explicit exception types over a
  broad `except Exception` so real I/O problems still surface. If a corruption mode is
  missing, users hit the old permanent-failure behaviour for that case.

Yes — 10 new tests, and I verified each one fails without the fix:

| Reverted to | Failing tests |
|---|---|
| plain `urlretrieve` (no temp file) | `test_failure_leaves_nothing_behind[OSError]`, `[KeyboardInterrupt]` |
| no recovery path | all 3 `test_recovers_from_corrupt_cache` params, `test_corrupt_cache_reports_clearly_when_download_disabled` |

`TestUrlretrieveAtomic` covers: failure leaves no residue (parametrized over `OSError`
and `KeyboardInterrupt`), failure preserves an existing cached file, the destination
does not exist *while the download is in flight* — the actual invariant a concurrent
reader depends on — and the promoted file's permissions.

`TestDownloadDatasetCache` covers recovery from empty / truncated / non-gzip entries,
that the repaired entry is then served from cache without a second download, and both
`download_if_missing=False` paths.

No network access: `ur.urlretrieve` is monkeypatched throughout.

`src/skfolio/datasets/_base.py` is now at **100% coverage** (82 statements, 0 missed).

#### Any other comments?

Full suite, two consecutive runs, macOS arm64 / Python 3.14.6, latest deps:

| | result |
|---|---|
| before | `1 failed, 6930 passed, 29 errors` |
| after | `1 failed, 6968 passed, 0 errors` (both runs) |

The remaining failure is `test_cvar_variance_views[TNC]` — pre-existing, unrelated,
tracked as #277.

The dataset suite also dropped from **24–38s to 1.3s**, because it no longer destroys
the cache it depends on and then re-downloads it. Across six earlier local runs the
error count had swung 28 → 3 → 2 → 1 → 0 → 29; it is now stably 0.


HugoDelatte:
Thanks for the PR.

I’ve simplified the implementation to focus on atomic downloads and test isolation:

- Download into a temporary directory inside `data_home`, then publish with `os.rename`.
- Handle `FileExistsError` when another Windows worker has already populated the cache.
- Let the temporary-directory context handle cleanup, including `KeyboardInterrupt`.
- Remove the separate helper, automatic corrupt-cache recovery, and forced `0644` permissions.
- Keep cache-deletion tests isolated from the user’s actual cache.

Added a deterministic concurrent-download test with the completed cache held open by a reader.

HugoDelatte (approved):

PR #283, opened 2026-09-01
test(docs): execute and verify the docstring examples

### LINKED ISSUE #275: test(docs): 945 of 1141 doctest examples are never executed
**Subcategory:** executable documentation — **4 → 9**

The codebase carries **1141 doctest examples** across 113 documented objects in 256
modules. Only **196 of them (17%) are ever executed.**

**File:** `pyproject.toml` `[tool.pytest.ini_options]`

```toml
addopts = ["--doctest-modules", ...]
testpaths = ["tests", "src/skfolio/distribution"]
```

`--doctest-modules` is enabled, but `testpaths` restricts collection to `tests` and the
single `src/skfolio/distribution` subpackage. The remaining **945 examples across the
other 26 subpackages are never run** — so a docstring example can silently stop matching
its output and nothing in the repo will notice.

**This is not hypothetical.** A malformed doctest already sits undetected in the tree:

`src/skfolio/pre_selection/_select_non_expiring.py:76` and `:86`

```
    ...)
```

The continuation marker requires `... ` followed by a space (as every other line in that
example correctly uses, e.g. `...    {`). Written as `...)` the example cannot be parsed
by `doctest` at all — reported as *"lacks blank after ...: '...)'"*. Because
`pre_selection` is outside `testpaths`, no test can see it.

Separately, `README.rst` contains **31 `.. code-block:: python` examples that are
executed by nothing** — Sphinx does not run `code-block`, and `tests/test_readme_shared_sections.py`
only guards the region markers, not example correctness.

**Suggested fix:** extend `testpaths` to `src` (or add a dedicated doctest session over
the full source root), fix the two malformed lines, and consider `--doctest-glob` over
`README.rst`.

**done when…** `pytest --doctest-modules src` collects and passes all 1141 examples, and
`_select_non_expiring.py` parses as a valid doctest.

tschm:
#### Reference Issues/PRs

Fixes #275. Part of the assessment tracked in #271.

`--doctest-modules` was already enabled, but `testpaths` scoped it to `tests` and
`src/skfolio/distribution`, so **only 196 of 1156 examples ever ran**. The rest are
rendered on skfolio.org and were never checked. Extending `testpaths` to `src` made
them run, and a number of them did not.

**Docstrings that failed to even parse** — the collector could not see their examples:

- `pre_selection/_select_non_expiring.py` had two `...)` continuation lines with no
  space after the ellipsis, hiding all 8 examples in that docstring.
- `model_selection/_walk_forward.py` prefixed three continuation lines with `>>> `,
  producing `IndentationError: expected an indented block after 'for' statement`.
- `optimization/cluster/hierarchical/_schur.py` was missing a closing parenthesis in
  the "advanced model" example, so it never constructed the estimator it documents.
  Its advanced weights were a verbatim copy of the default ones; they are now the
  real Ledoit-Wolf / Kendall / Ward output.

**Examples that could not run as written:**

- Three moments estimators did `X_nan[:50, 2] = np.nan` on a **DataFrame** — that is
  label-based indexing and raises `KeyError`. Now `.iloc[:50, 2]`.
- Both `CovarianceForecastEvaluation` and `CovarianceForecastComparison` referenced an
  `X` that the example never defines. They now load it like every other example.
- `moments/covariance/_base.py` printed values derived from an **unseeded**
  `np.random.randn`, so its documented output could never be reproduced.

**Output that was simply wrong:**

- `MultipleRandomizedCV` documented `Assets: columns=[0 1 4]` in all three examples,
  where `sample_unique_subsets(n=5, k=3, random_state=0)` returns `[1 2 3]`.
  `random.Random(0).sample` gives the same answer on 3.10 – 3.14, so these are now
  pinned to what the code actually produces.
- ~90 examples documented no output where the call in fact returns something, most
  often `>>> model.fit(X)` returning an estimator repr.

**Two leaked globals had to be neutralised**, because they change how documented
output renders and a reader of the published docs has neither:

- `tests/conftest.py:24` sets `np.set_printoptions(suppress=True, precision=6)` for
  the whole session. Harmless while doctests did not run over `src`; now it made
  arrays render at 6 decimals instead of numpy's 8.
- Several test modules call `sklearn.set_config(...)` without restoring it.

An autouse fixture in `src/skfolio/conftest.py` runs every doctest under the defaults.

#### How unstable values are documented

Values that are not bit-stable are written as the **actual output prefix** followed by
`...`, at three significant digits and keeping the scientific notation numpy uses:

```pycon
>>> print(portfolio.cvar)
0.0251...
>>> print(model.weights_)
[2.60...e-02 ... 6.89...e-03 7.15...e-02 ... 1.97...e-01 ... 1.20...e-01
 3.00...e-02 2.10...e-02 1.00...e-02 1.13...e-01 1.17...e-02 1.65...e-01
 9.86...e-03 9.69...e-03 1.15...e-01 8.92...e-02]
```

A bare `...` stands for a weight that is zero up to solver tolerance (`4.66e-12`),
whose digits are not reproducible across BLAS and solver versions. Long nested reprs
are matched as `MeanRisk(...)`; short, informative ones are kept in full.

#### Results

```
examples executed and verified   196 -> 1065
skipped, with reasons              0 ->   91  (7 docstrings)
docstrings that fail to parse      3 ->    0
```

Full suite: `7064 passed, 7 skipped`, identical over two consecutive runs.

- **The 7 skipped docstrings** in `src/skfolio/conftest.py`. Four are illustrative
  fragments referencing names that are never defined (`factor_returns`, `population`)
  and would need invented example data. The other three **raise**, and look like real
  bugs worth their own issue: `EWSharpeOptimalAlpha` and `PredictorAlpha` both
  document a `partial_fit(X[-5:])` call that re-feeds rows already seen
  (`ValueError: observations must be unique`), and `OpinionPooling`'s documented
  configuration does not converge (`SolverError: Solver 'CLARABEL' failed`). Happy to
  open that separately.
- **Whether three significant digits is enough margin on your CI matrix.** It removes
  the `e-12` noise, but a genuinely degenerate optimum could still differ in the third
  digit on a different BLAS. This is the one thing I cannot verify from a single
  machine — if a platform flakes, the fix is a shorter prefix at that one site.
- **`OnlineRandomizedSearch.best_params_`** is now documented. It is sampled from
  `scipy.stats.uniform` under `random_state=42`, so it is reproducible, but it is the
  entry most exposed to a scipy sampling change.

No new test files — the change *is* test coverage: 869 additional examples now execute
and have their documented output verified on every run. `src/skfolio/conftest.py` is
new and holds the skip list plus the global-state fixture.


tschm:
@HugoDelatte I am adding and(!) testing docstrings. Very helpful for all documentation purposes

tschm:
@HugoDelatte We make sure your code has docstrings which are also tested during ci/cd. A broken doctoring is just evil

HugoDelatte (changes requested):
Thanks for the PR!

My main concern is keeping the docstring examples easy to read.

I’d prefer:

```pycon
>>> print(portfolio.cvar)
0.0251...
```

over adding `.round(...)` or f-string precision just to satisfy doctests.

I think we should do these few changes:

- **Keep numerical examples simple.** Use actual output prefixes with `...`, preserving scientific notation where NumPy uses it. For example, `0.869556...` should become `0.869...`, not `0.8696...`. Keep short arrays complete.
- **Shorten repeated configurations.** Use `MeanRisk(...)`, `RiskBudgeting(...)`, etc. for long nested representations. Keep short, useful representations and labels.
- **Show the useful results.** Fill in missing output after `variance_[:5]`, `best_params_`, and `summary()`, using compact examples.
- **Remove global `NUMBER`.** It currently lets expected `[0.]` match actual `[0.75]`, which is too permissive for weights. Regenerate the expected outputs when removing it.
- **Keep `src/skfolio/conftest.py`, but simplify it.** One small fixture for the temporary directory and NumPy/sklearn output settings, plus one dictionary mapping skipped doctests to specific reasons, should be enough. The category sets, dictionary merges, and lengthy explanations seem unnecessary.
- **Reduce the quarantine through straightforward fixes.** Fix the Schur example’s missing parenthesis and regenerate its advanced weights, add the missing `X` setup, and complete deterministic examples such as `EmpiricalVariance`. Seed synthetic sampling explicitly where numerical results are shown.

Also update the quarantine rationale: Sphinx hides doctest flags by default, so avoiding visible markers in the published docs isn’t a reason to skip whole docstrings.

tschm:
Thanks @HugoDelatte — all six points are in, plus the rationale fix. Rundown:

**Removed the global `NUMBER`.** You were right that it was too permissive; every expected output is regenerated without it.

**Numerical examples now use actual output prefixes.** `.round(...)` and f-string precision are gone. Three significant digits, keeping numpy's scientific notation:

```pycon
>>> print(portfolio.cvar)
0.0251...
>>> print(model.regime_multiplier_)
0.869...
```

One thing to sanity-check: for the convex optimizers the raw array is genuinely scientific, because a handful of weights are solver noise:

```pycon
>>> print(model.weights_)
[2.60...e-02 ... 6.89...e-03 7.15...e-02 ... 1.97...e-01 ... 1.20...e-01
 3.00...e-02 2.10...e-02 1.00...e-02 1.13...e-01 1.17...e-02 1.65...e-01
 9.86...e-03 9.69...e-03 1.15...e-01 8.92...e-02]
```

The bare `...` is a weight that is zero up to solver tolerance (`4.66e-12` here) — its digits are not reproducible, so there is nothing honest to write. Everything else is a real prefix. If you would rather see the elided-middle style you use in `EntropyPooling` (`[0.073... 0.0541... ... 0.200...]`) at those sites, say the word.

**Shortened the long reprs** to `MeanRisk(...)`, `RiskBudgeting(...)`, `NestedClustersOptimization(...)`, `OpinionPooling(...)`, `TimeSeriesFactorModel(...)`. The short informative ones stay, e.g. `MeanRisk(linear_constraints=['AMD <= 0.10', 'BAC + JPM >= 0.15'], max_weights=0.2, risk_measure=CVaR)`.

**Filled in the missing results:** `variance_[:5]`, both searches' `best_params_`, all four `summary()` frames, and the six `CharacteristicsFactorModel` diagnostics.

**`conftest.py` is down to one `SKIPPED` dict and one fixture.** No category sets, no merges. The rationale no longer mentions rendered markers.

**Quarantine: 18 docstrings → 7.** The Schur parenthesis is fixed and its advanced weights regenerated — they were a verbatim copy of the default ones, so that example had never produced its own output. Both covariance-forecast examples got the `X` setup they referenced. `EmpiricalVariance` is complete. Vine-copula and synthetic-characteristics sampling are seeded explicitly.

While regenerating I also found that `MultipleRandomizedCV` documents `Assets: columns=[0 1 4]` in all three examples, where `sample_unique_subsets(n=5, k=3, random_state=0)` actually returns `[1 2 3]`. I checked `random.Random(0).sample` on 3.10–3.14 and it is stable, so those are pinned to the real values now.

The 7 that remain: four illustrative fragments (`factor_returns`, `population` are never defined — making them run means inventing example data, which is your call), and three that genuinely raise. Those three look like real bugs: `EWSharpeOptimalAlpha` and `PredictorAlpha` both document `partial_fit(X[-5:])` on rows already seen (`ValueError: observations must be unique`), and `OpinionPooling`'s documented configuration hits `SolverError: Solver 'CLARABEL' failed`. Happy to open an issue for them.

**Result:** 1065 of 1156 examples now execute and are verified, up from 196. Full suite `7064 passed, 7 skipped`, identical over two consecutive runs.

tschm:
Opened #292 for the raising examples — and I have to correct one thing I wrote above.

**`OpinionPooling` does not raise.** The pooling fits fine; it is the `RiskBudgeting` step further down the same example that fails. I had attributed the `SolverError` to the wrong call. Narrowed down:

| configuration | result |
| --- | --- |
| `RiskBudgeting(CVAR)` on the default prior | ok |
| `RiskBudgeting(CVAR)` on a single `EntropyPooling(cvar_views=["AMD == 0.10"])` | ok |
| `RiskBudgeting(CVAR)` on the **pooled** distribution | `SolverError: Solver 'CLARABEL' failed` |
| `RiskBudgeting(VARIANCE)` on the **pooled** distribution | ok |

The pooled weights span `1.17e-04` to `8.14e-03` with only 371 distinct values across 8312 scenarios — a much more skewed weighting than any single opinion produces, so this reads more like an ill-conditioned CVaR problem than a bug in the pooling. The skip reason in `conftest.py` now says this, and the `sample_weight` output the example *does* produce is filled in.

The two alpha ones are straightforward: `fit` retains a tail buffer, `partial_fit` prepends it, so `partial_fit(X[-5:])` re-feeds rows already held and `AssetPanel` rejects the duplicated observations. #292 has full repros for all three; whether `partial_fit` should tolerate overlap or the examples should feed genuinely new rows is your call.

---

Separately, 3.14 caught one of my own prefixes. `effective_number_of_scenarios_` was left at six significant digits (`6878.97...`); 3.14 returns `6879.324746471757` from the same solve. That is 5e-5 relative drift, so I had simply not applied my own three-significant-digit rule there — `687...` now. Every other platform in the matrix was green, including the min-deps job, so this was the single site where I kept more precision than the value supports. Fix pushed.

tschm:
3.14 got one line further and stopped at the entropy-pooling sample weights: `8.76...e-05` here, `8.75983160e-05` there. That value sits 2e-5 (relative) below a truncation boundary — a knife edge. Both entropy-pooling arrays are now at two significant digits.

Rather than keep discovering these one CI run at a time, I audited the whole diff: paired each of the 388 truncated prefixes with the value the run actually produces, and measured how far it sits from its own truncation boundary.

Seven others land within 5% of one:

```
0.009  -3.026...  <- -3.026991     CharacteristicsFactorModel
0.010  -0.331...  <- -0.331990     CharacteristicsFactorModel
0.023   0.006...  <- 0.006977      CharacteristicsFactorModel
0.030   0.000747... <- 0.00074797  EmpiricalVariance
0.031   7.74...   <- 7.74969306e-02  MaximumDiversification
0.036   2.91...   <- 2.91963931e-02  NestedClustersOptimization
0.048   0.0420... <- 0.04209519    RiskBudgeting
```

**I left all of them alone**, deliberately. Proximity to a boundary only matters in proportion to how much the value actually moves, and these differ by orders of magnitude:

- `CharacteristicsFactorModel`, `EmpiricalVariance`, `SchurComplementary` are deterministic linear algebra over a seeded panel — reproducible to roughly 1e-12.
- `MeanRisk` / `RiskBudgeting` / `MaximumDiversification` are convex solves at CLARABEL's default tolerance.
- Entropy pooling is an exponential-cone solve whose result is amplified through `exp()` — the only one here that demonstrably moves at 1e-4.

So a 1% margin on a value stable to 1e-12 is nine orders of margin, while a 2% margin on entropy pooling is a coin flip. All of these passed the full matrix — five Python versions, three operating systems, plus min-deps — twice; demoting them uniformly would throw away precision the platforms have already agreed on. My first instinct was to re-trim everything to a fixed relative margin, and that turned out to demote about a third of the diff to two digits for no evidence-backed reason.

If a site does flake later, the audit script tells us immediately which one and how much headroom it had, so the fix is a one-digit change rather than a hunt.

HugoDelatte (approved):
Looks good now.

I've just shortened the outputs and kept formatting in the doctest fixture, so examples use plain `print(...)` calls. Also fixed the alpha updates, PPF inputs and weight-cap example, and reduced the larger datasets to increase tests speed.

PR #284, opened 2026-09-02
chore(lint): bring examples under the ruff gate

tschm:
## Why

`examples/` sat outside `[tool.ruff] include`, so it had never been linted or formatted — 222 findings on first run.

## What

Added `examples/**/*.py` to the include list (this drives both `ruff check` and `ruff format`), plus a per-file-ignore for three rules that are structurally at odds with Sphinx-Gallery rather than actual defects:

| Rule | Why it's ignored for `examples/` |
|---|---|
| `D` | The module docstring is the RST page title; `D400`/`D205` would force periods and blank lines into gallery titles. |
| `E402` | Imports are deliberately interleaved with the `# %%` narrative cells. |
| `B018` | A bare trailing expression is how a cell renders a value in the gallery output. |

That accounts for 187 of the 222. The remaining 35 are real and fixed:

- **`RUF002`/`RUF003` ambiguous unicode.** The en dashes in `plot_5_weight_constraints.py` were `->` arrows mistyped as `–>` — the same bullet list already used plain `->` on other lines, so this is now internally consistent. Curly apostrophes → `'`.
- **`I001`** unsorted imports in 8 files.
- **`F401`** unused `numpy` in `plot_15_mip_cardinality_constraints.py`.
- **`RUF046`** in `plot_3_custom_pre_selection_volumes.py` — `int(round(x))` where `x` is a Python float, so the cast is a no-op.

`ruff format` then reflowed 7 files: long `train_test_split` calls, multi-import blocks, and blank lines around class definitions.

## Verification

Run against ruff 0.16.5 — the version `uv.lock` pins and CI actually runs:

- `ruff check` and `ruff format --check` are clean across the repo (507 files).
- All 47 examples compile.
- `plot_3_custom_pre_selection_volumes.py`, the only behavioural edit, runs to exit 0.

## Note, out of scope here

`.pre-commit-config.yaml` pins ruff at `v0.8.4` while `uv.lock` resolves 0.16.5. That hook already fails on 14 pre-existing `UP038`/`B008` findings in `src/` on `main`, independently of this PR — so it is currently out of sync with the CI gate. Probably worth a separate bump.


examples/clustering/plot_1_hrp_cvar.py
```diff
@@ -103,7 +105,7 @@
 # linkage method is set to the Ward variance minimization algorithm, which is more
 # stable and has better properties than the single-linkage method.
 #
-# However, since the HRP optimization doesn’t utilize the full cluster structure but
+# However, since the HRP optimization doesn't utilize the full cluster structure but
 # only their orders, the allocation remains relatively stable regardless of the chosen
```
tschm, line 109: I wouldn't call this behaviour stable :-) I think this route was taken back then as LdP wasn't aware of different linkage methods. He was using "simple" and had full exposure to its chaining effect.

HugoDelatte: fair point, happy to revisit defaults and wording in another PR if you think it's needed.

tschm:
@HugoDelatte That's an easy one. Just checking whether you respect your ruff rules also for your examples. This is then tested during ci/cd

HugoDelatte (approved):
Thanks for the PR, it's a good idea.
I've just included the examples in the pre-commit Ruff hooks too.

PR #285, opened 2026-09-02
fix(measures): apply the Bessel correction over non-NaN observations in semi_variance

WatchTree-19:
`semi_variance`'s unbiased branch divides by `len(returns)`, the raw row count, while its numerator goes through `nanmean`. When the input contains NaN the Bessel correction is computed over a larger sample than the one actually used, so the result is understated by roughly the NaN fraction.

`variance` does not have this problem because it delegates to `np.nanvar`, so the divisor is the non-NaN count. Both functions carry the same documented contract:

> NaN handling:
> - Unweighted: NaNs are ignored; all-NaN inputs yield NaN.

`semi_variance`'s biased branch honours it. The unbiased branch does not.

## Reproduction

Using the repo's own `returns_1d_nan` fixture, 4 valid observations in 7 rows:

```
biased semi_variance   : 0.09531250
unbiased semi_variance : 0.11119792   <- scaled by 7/6
correct                : 0.12708333   <- should scale by 4/3
understated by         : 12.5%

variance biased -> unbiased ratio: 1.3333 (4/3, correct)
```

Two further symptoms from the same line:

| input | `variance` | `semi_variance` (before) |
|---|---|---|
| `[nan, 0.1, nan]`, 1 valid | `nan` | **`0.0`** |
| `[0.1]`, n=1, no NaN | `nan` | **`ZeroDivisionError`** |

Returning `0.0` reads as "zero downside risk" rather than "not enough data".

The error is always negative, so downside risk is systematically understated and never overstated, and its magnitude tracks the NaN fraction: about 0.3% at 250 valid of 1000, 12.5% on the fixture above, 20% at 5 of 1000.

## The change

One hunk. Count non-NaN observations per column, and return NaN below two observations, which is what `np.nanvar` already does for `variance`. It is a strict no-op when there are no NaNs, and the `sample_weight` path is untouched.

## Scope

I checked how far this reaches before writing it up, because I would rather state the limit than oversell it.

- **`MeanRisk`, `RiskBudgeting`, `DistributionallyRobustCVaR` and `BenchmarkTracker` are unaffected.** They solve through `_semi_variance_risk` / `_semi_deviation_risk` in `optimization/convex/_base.py`, an independent CVXPY formulation. I confirmed weights are bit-identical before and after on the bundled S&P 500 dataset, max abs diff `0.000e+00`.
- **`Portfolio` on clean data is unaffected** — identical `semi_variance`, `semi_deviation` and `sortino_ratio`.
- The reachable paths are the functional API (`skfolio.measures.semi_variance` on a returns panel with missing data, which is the normal shape of a mixed-listing-date universe) and the hierarchical optimisers (HRP, HERC, NCO), which read `getattr(ptf, risk_measure.value)` with `biased=False`.

So this is a correctness fix in the measures module, not an optimiser-corrupting defect.

## Tests

All 258 existing tests pass unchanged.

Every existing NaN case for these two measures used `biased=True`, which returns before the Bessel correction, so the one branch that differs from `variance` was never exercised with NaN. That is why it survived. Added:

- unbiased `1d_nan`, `2d_nan` and `all_nan` cases for both measures, plus the missing biased `2d_nan` case that `test_variance` already has
- an invariant test: padding a series with NaN must not change the result
- a small-sample test: fewer than two non-NaN observations yields NaN, matching `variance`

270 pass with the fix. 8 of the new cases fail without it.

## One thing I did not change

`optimization/convex/_base.py:1829` computes `cp.sum_squares(v) / (n_observations - 1)` on the raw row count, which is the same idea in the CVXPY formulation. I left it alone: CVXPY will not accept NaN input, so it is not reachable the same way, and it felt out of scope here. Happy to look at it separately if you think it is worth doing.


HugoDelatte (approved):
Thanks for the fix. I agree with it.

I made a small cleanup to reuse the existing `safe_divide` helper, make the observation count clearer, and shorten the comments.

PR #286, opened 2026-09-03
[ENH] Add weight drift

### LINKED ISSUE #161: [ENH] Add optional TC and weight drift in `cross_val_predict`
* Add optional  TC and  weight drift in `cross_val_predict`
* Add a doc section that explain the pros/cons for using non-compounded vs compounded returns / no weight drift vs  weight drift / smooth TC vs exact TC --> in term of unbiased expectation of cumulative simple return (e.g. measuring strategy "skill" or “alpha”) vs expected terminal wealth vs  single sample (historical) path.

HugoDelatte:
Fixes #161

Add `weight_drift` to evaluate portfolios with drifted weights.

By default, a `Portfolio` holds its target weights on every observation. This PR adds
`weight_drift` to evaluate the same portfolio with weights that drift with the asset
returns over the observation window, and wires it into sequential cross-validation.

- `Portfolio(weight_drift=True)`: the held weights follow the self-financing identity
  from `weights` on the first observation and the returns are those of the drifted
  weights. The implicit cash position earns zero. Transaction costs, management fees
  and `compounded` behave as before.
- New `Portfolio.final_drifted_weights`, `final_drifted_weights_dict` and `turnover`.
  `weights_per_observation` and realized attribution use the drifted weights when
  `weight_drift=True`.
- New `MultiPeriodPortfolio.final_drifted_weights_dict`, `turnover` and `total_cost`.
- `cross_val_predict` and `online_predict`: `portfolio_params["weight_drift"]` is
  forwarded to each `Portfolio` of the path, the other parameters configure the
  returned `MultiPeriodPortfolio`. Setting it makes the path sequential and passes the
  end-of-period drifted weights as `previous_weights` to the next fit, so
  `max_turnover` and `transaction_costs` are measured against the held weights instead
  of the previous targets.

- Docs: new Evaluation Conventions user guide covering weight drift, compounding and
  transaction cost, with the bias and variance of each convention and the
  settings to use for each evaluation goal.


docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+**************************
+
+Portfolio backtests serve several distinct objectives. They can estimate allocation
+skill, compare portfolio strategies, quantify implementation effects or reconstruct a
+historical wealth path. Each objective may require a different construction of
+portfolio returns and performance statistics.
+
+`skfolio` is designed for the statistical evaluation of portfolio strategies: the
+observed history is one realization of the process generating returns. Statistics
+estimated from it can be biased and carry sampling uncertainty.
```
FarrellDaniel: "and carry sampling uncertainty" 

better for another verb - e.g. "and will carry uncertainty due to the sampling process" etc.

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+Backtesting and Evaluation
+**************************
+
+Portfolio backtests serve several distinct objectives. They can estimate allocation
+skill, compare portfolio strategies, quantify implementation effects or reconstruct a
+historical wealth path. Each objective may require a different construction of
+portfolio returns and performance statistics.
+
+`skfolio` is designed for the statistical evaluation of portfolio strategies: the
+observed history is one realization of the process generating returns. Statistics
```
FarrellDaniel: use of semicolon is AI and not sure it reads well - I think this whole paragraph hangs a bit and probably belongs merged into the first paragraph instead

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+estimated from it can be biased and carry sampling uncertainty.
+
+The evaluation should distinguish four objectives:
```
FarrellDaniel: "The evaluation" is introduced from nowhere

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+
+* **Allocation skill (ex-ante).** Estimate the expected return and risk of the
+  target allocations selected by the optimizer, isolating the allocation decisions
```
FarrellDaniel: optimizer --> optimization process?

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+  from the subsequent drift of the holdings.
```
FarrellDaniel: is this the actual definition of "allocation skill" --> drift of the holdings means what exactly? hasn't been properly introduced yet

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+* **Expected performance of the implemented strategy.** Estimate the expected
+  return and risk of rebalancing to the targets and letting holdings drift between
```
FarrellDaniel: "to the target weights, allowing holdings to drift between rebalance periods"

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+  rebalances. Turnover and transaction costs reflect the trades needed to move from
+  the drifted holdings to the new target weights.
```
FarrellDaniel: "drifted holdings" doesn't sound great --> "reflect the effects of trades needed to move towards the new target weights"

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+* **Relative strategy performance.** Estimate a difference in expected performance.
+  Common biases may cancel, and positively correlated sampling errors can reduce
+  uncertainty in a paired comparison on the same out-of-sample dates.
+* **Historical realization (ex-post).** Reconstruct the holdings, trades, wealth
+  and drawdowns along the observed return path under stated implementation
+  assumptions.
+
+An accurate ex-post reconstruction does not by itself establish what to expect on a
+new path. The distinction is between reconstructing an outcome and drawing statistical
```
FarrellDaniel: new path means 'out of sample'?

"The distinction" refers to what?

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+The evaluation should distinguish four objectives:
+
+* **Allocation skill (ex-ante).** Estimate the expected return and risk of the
+  target allocations selected by the optimizer, isolating the allocation decisions
+  from the subsequent drift of the holdings.
+* **Expected performance of the implemented strategy.** Estimate the expected
+  return and risk of rebalancing to the targets and letting holdings drift between
+  rebalances. Turnover and transaction costs reflect the trades needed to move from
+  the drifted holdings to the new target weights.
+* **Relative strategy performance.** Estimate a difference in expected performance.
```
FarrellDaniel: between two strategies?

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+* **Historical realization (ex-post).** Reconstruct the holdings, trades, wealth
+  and drawdowns along the observed return path under stated implementation
+  assumptions.
+
+An accurate ex-post reconstruction does not by itself establish what to expect on a
+new path. The distinction is between reconstructing an outcome and drawing statistical
+conclusions from it.
+
+These objectives guide how portfolio returns are constructed and accumulated,
+and how the resulting statistics are interpreted. Two parameters govern return
```
FarrellDaniel: Would move this above to motivate why you are introducing the four objectives in the first place. Then new paragraph, "With respect to return construction and accumulation, two parameters typically govern the modelling choices available:"

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+  return. With `False`, returns are computed at the **target weights (no drift)**,
+  which remain constant within each holding period. With `True`, returns are
+  computed from **drifted weights** as asset prices change between rebalancing dates.
+* `compounded` determines how those returns are accumulated. With `False`,
+  cumulative returns are an arithmetic sum. With `True`, they form a compounded
+  wealth index. This setting also affects drawdown measures, but leaves the
+  underlying return series unchanged.
+
+Both default to `False`. The default evaluates **allocation skill (ex-ante)** at
+target weights, consistent with the optimizer's linear portfolio return definition.
```
FarrellDaniel: ...return definition [and] cumulative returns are...

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+   * - Expected performance of the implemented strategy
+     - `True`: drifted weights
+     - `False`
+   * - Relative strategy performance
+     - `False`: target weights (no drift). `True`: drifted weights.
+     - `False`
+   * - Historical realization (ex-post): wealth and percentage drawdowns
+     - `True`: drifted weights
+     - `True`
+
```
FarrellDaniel: does this mean you would run 3 versions with each of the combinations? [(false, false), (true, false), (true, true)]? how do you disentangle allocation skill from relative performance if they are both the same?

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+conclusions from it.
+
+These objectives guide how portfolio returns are constructed and accumulated,
+and how the resulting statistics are interpreted. Two parameters govern return
+construction and accumulation:
+
+* `weight_drift` determines the weights used to compute each observation's
+  return. With `False`, returns are computed at the **target weights (no drift)**,
+  which remain constant within each holding period. With `True`, returns are
+  computed from **drifted weights** as asset prices change between rebalancing dates.
```
FarrellDaniel: and in the case of false, would a transaction cost be applied to the implicitly trade needed to rebalance back to the target weight?

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+
+`compounded` changes cumulative returns, drawdown measures and ratios based on
+drawdowns. Measures computed from individual returns, such as mean return, variance,
+VaR, CVaR, and the Sharpe and Sortino ratios, are unchanged, as are their sampling
+variances.
+
+`weight_drift` changes the underlying returns and can change both the bias and
+sampling variance of performance estimates for the chosen evaluation objective.
+
+Reconstructing holdings and trades more accurately does not necessarily improve
```
FarrellDaniel: "more accurately" refers to "weight drift" = False, correct? why do you label this as more accurate than the other...they are both valid choices which you can implement in the real world...?

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+estimates of expected performance. An evaluation at target weights can have lower
+mean squared error if its reduction in sampling variance outweighs its squared bias.
+See :ref:`weight_drift_mse`.
+
+The figure below follows two illustrative ten-year paths from the same asset return
+model, with equal target weights and annual rebalancing. It adds weight drift and
+then compounding while keeping the target allocations fixed. The paths were selected
+to illustrate the differences. The gap is not an estimate of the typical effect of
```
FarrellDaniel: "not meant to provide an estimate"

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+   \qquad
+   u_{t+1} = \frac{u_t \circ (1 + r_t)}{1 + u_t \cdot r_t},
+   \qquad u_{t_k} = w_k,
+
+where :math:`\circ` denotes element-wise multiplication. The weights reset to
+the next targets at :math:`t_{k+1}`.
+
+For example, start with 50% in each of two assets. If the first asset gains 10%
+and the second is unchanged, the portfolio gains 5%. The next observation starts
+with weights of 52.38% and 47.62%. An evaluation at target weights uses 50% and 50%
```
FarrellDaniel: ...whereas an evaluation at target weights uses 50% and 50% instead.

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+again.
+
+If the first asset outperforms again, the drifted portfolio has a higher return
+than the target-weight portfolio. If their relative performance reverses, it has
+a lower return. Momentum and reversal can therefore affect the comparison.
```
FarrellDaniel: "comparison" --> do we mean "can affect the performance of the strategy under the two modelling assumptions"?

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+
+The drift calculation assumes zero interest on cash.
```
FarrellDaniel: I think this needs some qualification...also in a drifted portfolio I wouldn't expect there to be any realised cash (you keep everything invested in the assets and let the weights move...whereas in a drift = False you are implicitly trading and therefore a properly accurate replication would introduce cost of cash on the realised pnl?).

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+
+`Portfolio` raises an error when wealth becomes non-positive.
```
FarrellDaniel: perhaps qualify when this can happen?

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+* **Allocation skill (ex-ante).** Before transaction costs, `weight_drift=False`
+  introduces no bias from drift. `weight_drift=True` introduces bias :math:`\bar d`.
+* **Expected performance of the implemented strategy.** For expected net return,
+  `weight_drift=True` accounts for drift and executed costs. `weight_drift=False`
+  introduces bias :math:`b-\bar d`.
+
+The expected gross-return difference between the two settings can be negligible
+with frequent rebalancing. In illustrative examples, it is about **0.2 bp p.a.**
+for a 50-asset equity portfolio rebalanced weekly, and about **2% p.a.** for a
+four-asset crypto portfolio rebalanced quarterly.
```
FarrellDaniel: first time we introduce crypto into the docs? do we want to...? do we actually evidence these meaning we've added datasets? Is it perhaps better to say "high volatility assets, e.g. small cap or crypto assets"?

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+These examples use equal target weights and geometric Brownian asset prices.
```
FarrellDaniel: so they're simulated...meaning...why do we label them as crypto?

docs/user_guide/portfolio.rst
```diff
@@ -111,26 +112,26 @@ returns, :math:`w` the vector of assets weights, :math:`c` the vector of assets
...
-    can dominate portfolio results because it moved early, not because it has a higher
-    expected return.
+By default, the target weights are held on every observation. This constant-weight
+convention is consistent with the convex optimization problems: it evaluates the target
+allocation the optimizer selected. It measures **allocation skill**, an
+expectation-based (ex-ante) evaluation whose return distribution does not depend on the
+order of the observations.
+
+With `weight_drift=True`, each position grows with its own asset return and the held
+weights drift away from the targets between the first and the last observation of `X`
```
FarrellDaniel: and the ~held~ weights drift away

docs/user_guide/portfolio.rst
```diff
@@ -111,26 +112,26 @@ returns, :math:`w` the vector of assets weights, :math:`c` the vector of assets
...
+(buy-and-hold). The implicit cash position earns zero. Early winners take a larger
```
FarrellDaniel: "implicit cash position" raises more questions without further qualification...

FarrellDaniel: "Early winners and losers" could be rephrased more neutrally as e.g. "Assets that experience an immediate increase in value will therefore inherit a larger weighting relative to those that have decreased in value (and all with respect to their initial target weight), such that the return series depends on the order of the observations."

docs/user_guide/portfolio.rst
```diff
@@ -111,26 +112,26 @@ returns, :math:`w` the vector of assets weights, :math:`c` the vector of assets
...
+weight and early losers a smaller one, so the return series depends on the order of the
+observations. Combined with `compounded=True`, it measures **realized capital growth**,
+a path-dependent (ex-post) evaluation along the historical return path.
+
+Both weight conventions use the same transaction-cost and management-fee formulas.
+In a sequential evaluation, drift can still change transaction costs at later
+rebalancing dates because one portfolio's `ending_weights` become the next portfolio's
+`previous_weights`. Management fees remain based on the target weights.
```
FarrellDaniel: management fees can remain on target weights because they are taken universally across the entire portfolio, the weights of which always sum to 100% in both drift and non-drift, correct?

src/skfolio/portfolio/_portfolio.py
```diff
@@ -43,24 +42,24 @@ class Portfolio(BasePortfolio):
...
+    By default, the portfolio returns are the dot product of the target `weights` and
+    the asset returns, minus transaction costs and management fees. The target weights
+    are held on every observation of `X`, consistent with the convex optimization
+    problems where the weights are the decision variables. This constant-weight
+    evaluation (`weight_drift=False`) measures **allocation skill**: an
+    expectation-based (ex-ante) evaluation whose return distribution does not depend
+    on the order of the observations.
+
+    With `weight_drift=True`, each position grows with its own asset return and the held
+    weights drift away from the targets over the window of `X` (buy-and-hold). This is
```
FarrellDaniel: "held weights"

src/skfolio/portfolio/_portfolio.py
```diff
@@ -178,6 +179,18 @@ class Portfolio(BasePortfolio):
...
 
+    weight_drift : bool, default=False
+        If this is set to True, the weights held on each observation drift with the
+        asset returns following the self-financing identity
+        :math:`u_{t+1} = u_t \circ (1 + r_t) / (1 + u_t \cdot r_t)`, starting from
+        `weights` on the first observation of `X`. The portfolio returns are those of
+        the drifted weights. Drift accumulates over the whole window of
+        `X`, and the implicit cash position :math:`1 - \sum_i w_i` earns zero.
+        The same transaction-cost and management-fee formulas are used with either
+        setting. The default (`False`) is to hold the target `weights` on every
```
FarrellDaniel: why is it 'weights'?

and the implicit cash position is only because weights sum to less than 1, not because of a choice of either drift or non-drifted...?

src/skfolio/portfolio/_portfolio.py
```diff
@@ -779,17 +857,78 @@ def previous_weights_dict(self) -> dict[str, float]:
...
+    @property
+    def turnover(self) -> float:
+        """Total absolute weight traded at the start of the period.
+
+        In a sequential evaluation, `previous_weights` come from the last successful
+        Portfolio. With `weight_drift=False`, they are its target weights, so this is
+        target turnover. With `weight_drift=True`, they include the intervening drift,
+        so this is executed turnover. When `previous_weights` is None, it defaults to
+        zero. Turnover includes the full absolute weight of positions in assets
+        absent from `X`.
```
FarrellDaniel: potentially something more verbose to define target turnover and executed turnover, such as: 

"With weight_drift=False, previous_weights are the previous Portfolio's target weights, so turnover measures the change between successive target allocations. With weight_drift=True, previous_weights are the weights actually held after returns have caused them to drift, so turnover measures the trades required to rebalance those holdings to the new target."

docs/user_guide/backtesting_and_evaluation.rst
```diff
@@ -0,0 +1,496 @@
...
+.. math::
+
+   \text{target turnover} = \|w_{k+1}-w_k\|_1,
+   \qquad
+   \text{executed turnover} = \|w_{k+1}-\tilde w_k\|_1.
+
+In a sequential evaluation, `weight_drift=False` uses target turnover for
+transaction costs, and `weight_drift=True` uses executed turnover. Target turnover
+measures changes in allocation decisions. Executed turnover also reflects price
+changes since the previous trade.
```
FarrellDaniel: something slightly more verbose: 

With weight_drift=False, previous_weights are the previous Portfolio's target weights, so turnover measures the change between successive target allocations. With weight_drift=True, previous_weights are the weights actually held after returns have caused them to drift, so turnover measures the trades required to rebalance those holdings to the new target.

FarrellDaniel (approved):

PR #290, opened 2026-09-08
test(prior): stabilize entropy pooling tests and update CI to Python 3.14

### LINKED ISSUE #277: test(prior): test_cvar_variance_views[TNC] fails deterministically on Python 3.14

HugoDelatte:
Fixes #277

Stabilize entropy pooling tests across numerical environments:

- Use `ftol=1e-10` in `test_cvar_variance_views[TNC]`, preserving all output-accuracy assertions.
- Check kurtosis inequalities in their requested direction, allowing 0.1% numerical error at the BAC boundary.
- Set `MAIN_PYTHON` to 3.14, covering the previously untested macOS ARM/Python 3.14 combination without adding CI jobs.


tschm:
@HugoDelatte please merge. I will update the other merge candidates

FarrellDaniel (approved):

PR #291, opened 2026-09-08
test: make the suite hermetic and scope sklearn config

### LINKED ISSUE #278: test: make the suite hermetic — network fixtures and leaked sklearn config
**Subcategory:** test design quality — **8 → 10**

Two independent sources of non-determinism in an otherwise strong suite (6780
assertions, 291 `parametrize` decorators, 883 `pytest.raises`).

**1. The suite is not hermetic — it downloads over the network at test time.**

`tests/conftest.py:57`

```python
@pytest.fixture(scope="module")
def implied_vol():
    implied_vol = load_sp500_implied_vol_dataset()
```

`load_sp500_implied_vol_dataset()` fetches from
`https://github.com/skfolio/skfolio-datasets/raw/main/datasets/...` on first use. Tests
therefore fail offline, and are exposed to the cache-corruption race tracked separately
(see the `download_dataset` atomicity issue). Observed as 28, then 3, then 1 setup errors
across consecutive runs.

**2. Three modules mutate global sklearn config without restoring it.**

- `tests/test_pipeline/test_pipeline.py:31` — `set_config(transform_output="pandas")`
- `tests/test_pre_selection/test_select_k_extremes.py:11` — same
- `tests/test_pre_selection/test_select_non_dominated.py` — same

The rest of the suite correctly uses the scoped `with config_context(...)` form (20+
modules do). Under `-n=4 --dist=worksteal` the leaked setting persists for every
subsequent test on the same worker, and worksteal makes that ordering vary run to run.
I did not manage to pin a specific failure on this, but it is a latent source of exactly
the kind of order-dependent flake that is expensive to diagnose later.

**Suggested fix:** cache the datasets as test fixtures (or mark the network tests and
skip them offline), and convert the three `set_config` calls to `config_context` blocks
or an autouse restoring fixture.

**done when…** the full suite passes offline, and `grep -rn "set_config" tests/` returns
only scoped `config_context` usages.

tschm:
Closes #278.

## 1. The suite no longer downloads at test time

Three loaders fetch from `skfolio-datasets` on first use: `load_sp500_implied_vol_dataset`, `load_nasdaq_dataset` and `load_ftse100_dataset`. They are now reached through a session-scoped `remote_dataset` fixture that

- tries `download_if_missing=False` first, so a warm cache never touches the network;
- attempts a single download otherwise;
- `pytest.skip`s on `URLError`/`TimeoutError` rather than failing the test.

`pytest_collection_modifyitems` marks every item whose fixture closure contains `remote_dataset` with a new `network` marker (registered in `pyproject.toml`). That covers the 36 tests that reach it transitively through the `implied_vol` fixture and the `nasdaq_X` fixture in `test_stats.py`, plus the six loader tests in `test_dataset.py` — 44 in total — without a marker on each test, and keeps working as tests are added. `pytest -m "not network"` now gives a fully offline run.

## 2. Scoped sklearn config

`test_pipeline.py`, `test_select_k_extremes.py` and `test_select_non_dominated.py` called `set_config(transform_output="pandas")` at test level and never restored it, so the setting leaked into every later test on the same `-n=4 --dist=worksteal` worker in a run-dependent order. All three now use the `with config_context(...)` form the other 20+ modules already use; `grep -rn "set_config" tests/` returns nothing.

## Verification

- Offline (empty `SKFOLIO_DATA`, proxy pointed at a dead port): `6912 passed, 55 skipped`, exit 0.
- Online, unchanged config: `6967 passed`, exit 0.
- `ruff check` and `ruff format --check` clean.


tschm:
@HugoDelatte The setting I mentioned is 

Always suggest updating pull request branches Loading
Whenever there are new changes available in the base branch, present an “update branch” option in the pull request.

HugoDelatte:
@tschm great idea, I've added the option now.

HugoDelatte (approved):
Thanks for the PR!

Looks good overall, I just made several targeted improvements:

- Extended offline handling to the three remote dataset doctests, including network markers and shared skip handling.
- Removed the redundant cache-only loader call; the loaders already check their cache.
- Kept HTTP errors as failures, while connectivity errors and timeouts still skip.
- Replaced the large zero-filled bootstrap expectations with compact diagonal matrices and tighter relative tolerances, preserving off-diagonal checks.
- Corrected the non-diagonal bootstrap test to instantiate `BootstrapMuUncertaintySet` instead of `EmpiricalMuUncertaintySet`.
- Shortened the reproducibility test’s docstring and included the shared root fixture in Ruff and pre-commit checks.

PR #293, opened 2026-09-09
fix(optimization): retry a failed solve with a fallback solver

### LINKED ISSUE #292: OpinionPooling example raises: CVaR risk budgeting fails on the pooled distribution
`OpinionPooling`'s published example raises. The pooling itself fits fine — it is the
`RiskBudgeting` step in the same docstring that fails.

Reproduces on `main` (`b500f88`), Python 3.14.6, macOS.

`src/skfolio/prior/_opinion_pooling.py`:

```python
from skfolio import RiskMeasure
from skfolio.datasets import load_sp500_dataset
from skfolio.prior import EntropyPooling, OpinionPooling
from skfolio.optimization import RiskBudgeting

X = prices_to_returns(load_sp500_dataset())
opinion_pooling = OpinionPooling(
    estimators=[
        ("opinion_1", EntropyPooling(cvar_views=["AMD == 0.10"])),
        ("opinion_2", EntropyPooling(mean_views=["AMD >= BAC", "JPM <= prior(JPM) * 0.8"],
                                     cvar_views=["GE == 0.12"])),
    ],
    opinion_probabilities=[0.4, 0.5],
)
opinion_pooling.fit(X)   # ok

RiskBudgeting(risk_measure=RiskMeasure.CVAR, prior_estimator=opinion_pooling).fit(X)
# cvxpy.error.SolverError: Solver 'CLARABEL' failed.
```

What narrows it down:

| configuration | result |
| --- | --- |
| `RiskBudgeting(CVAR)` on the default prior | ok |
| `RiskBudgeting(CVAR)` on a single `EntropyPooling(cvar_views=["AMD == 0.10"])` | ok |
| `RiskBudgeting(CVAR)` on the **pooled** distribution | `SolverError` |
| `RiskBudgeting(VARIANCE)` on the **pooled** distribution | ok |

So it is the combination of CVaR risk budgeting with the pooled sample weights. Those
weights span `1.17e-04` to `8.14e-03` (a ~69x spread) with only 371 distinct values
across 8312 scenarios, which is a much more skewed weighting than any single opinion
produces — plausibly an ill-conditioned CVaR problem rather than a bug in the pooling.

I could not cross-check against a second conic solver: `solver="SCS"` raises
`ValueError: The solver SCS is not installed.` in this environment.

### Why it matters

This is a user-facing example on skfolio.org that has presumably never run. Anyone
copying it hits the error immediately. It stays in `SKIPPED` in
`src/skfolio/conftest.py` with the reason recorded, so it is verified the moment it is
fixed — it is now the only entry there that is a real failure rather than an
illustrative fragment.

---

<sub>Originally filed as three failures. The two `partial_fit` examples
(`EWSharpeOptimalAlpha`, `PredictorAlpha`, both `ValueError: observations must be
unique` on re-fed rows) were fixed in #283 by restructuring them to `fit(X[:-5])` then
`partial_fit(X[-5:])`; `_prepend_buffer` still rejects overlapping observations. Both
docstrings are out of `SKIPPED` and pass. This issue now tracks only the
`OpinionPooling` failure.</sub>

tschm:
#### Reference Issues/PRs

Fixes #292. The example was the last entry in `SKIPPED` from #283 that was a real
failure rather than an illustrative fragment.

> **Revised after review.** The first version of this PR added an automatic solver
> retry inside `ConvexOptimization`. Following @HugoDelatte's review that is fully
> reverted: **no library behaviour changes here.** The diff is now the example, the
> doctest configuration, the tests and the dependency groups.

`OpinionPooling`'s documented example raised:

```python
RiskBudgeting(risk_measure=RiskMeasure.CVAR, prior_estimator=opinion_pooling).fit(X)
# cvxpy.error.SolverError: Solver 'CLARABEL' failed.
```

**It is not a bug in the pooling.** It reproduces with a *single* `EntropyPooling`
opinion whose views concentrate `sample_weight` (`opinion_2` alone), so the issue's
"combination of CVaR risk budgeting with the pooled sample weights" is really *CVaR
risk budgeting with any concentrated `sample_weight`*.

**It is also a clarabel regression.** The minimum-dependency CI job caught this:
**clarabel 0.10.0, the declared minimum, solves both instances; only 0.11 stalls.**

The instance is well posed. Minimising CVaR over the simplex on that distribution
gives `+0.04163289`, so the problem is bounded and feasible, and the log-barrier
formulation is not unbounded. CLARABEL 0.11 simply stalls — the dual residual freezes at
`4.15e-02` with a step size of `0.00e+00` and it terminates in `InsufficientProgress`:

```
 12  +1.1321e+00  +1.1307e+00  1.22e-03  1.26e-04  4.15e-02  5.81e-05  4.98e-04  6.83e-01
 13  +1.1321e+00  +1.1307e+00  1.22e-03  1.26e-04  4.15e-02  5.81e-05  4.98e-04  0.00e+00
 14  +1.1321e+00  +1.1307e+00  1.22e-03  1.26e-04  4.15e-02  5.81e-05  4.98e-04  0.00e+00
Terminated with status = InsufficientProgress
```

#### The change

**SCS solves it.** It is a first-order method, so its failure modes are unrelated to
an interior point method's. Over 48 risk-measure × prior combinations the two solvers
agree wherever both succeed (to ~1e-4, consistent with SCS's lower default accuracy)
and **no combination fails both**.

The example reaches it through the existing estimator `fallback`, which keeps solver
selection explicit, preserves the prior, risk measure and constraints, and gives SCS
its own tolerances and iteration limit:

```python
>>> model = RiskBudgeting(
...     risk_measure=RiskMeasure.CVAR,
...     prior_estimator=opinion_pooling
... )
>>> model = model.set_params(
...     fallback=clone(model).set_params(
...         solver="SCS",
...         solver_params={"eps_abs": 1e-6, "eps_rel": 1e-6, "max_iters": 100_000},
...     )
... )
>>> model.fit(X)
```

`model = model.set_params(...)` rather than a bare `set_params` call only so the
doctest does not have to swallow the repr. The successful fallback and the failed
primary attempt are available through `fallback_` and `fallback_chain_`.

**`scs` is not a runtime dependency.** It joins the `dev` and `docs` groups, which is
how `PySCIPOpt` is already handled, and `install.rst` now names SCS in *Install
Additional Solvers* with the reason. That keeps the lean `cvxpy-base` + `clarabel`
install intact; the example documents `pip install scs` inline.

#### What I tried and rejected

None of these is in the diff. Each trades one silent failure for another, which is
worse than the current loud one:

| attempt | outcome |
| --- | --- |
| relax `tol_gap_abs`/`tol_gap_rel` to 1e-8, 1e-6, 1e-4; the `reduced_tol_*` family | never helps — the iterations **stall**, they do not stop short of an accuracy target |
| `scale_objective` × `scale_constraints` sweep | knife edge: `0.1/1` solves, `0.1/10` fails, `1/1` fails |
| anchor the log barrier at `1/n_assets` instead of `1` (provably solution neutral, since every risk measure is positively homogeneous and `weights_` is `w / factor`) | fixes both CVaR instances but turns cases that solve cleanly today into `optimal_inaccurate` or failures |
| encode the barrier as `cp.geo_mean(w, p=risk_budget) >= 1` (identical constraint, SOC instead of exponential cones) | fixes CVaR, **breaks EVaR**, and canonicalizes in quadratic time — 15s at 300 assets, >30s at 1000 |
| the same with `approx=False` (true power cones) | builds in 0.14s at 1000 assets but fails nearly everywhere, including `CVAR` on the default prior |
| `max_step_fraction=0.5` | fixes the documented example, not `opinion_2` |
| SCIP | also fails |

The `geo_mean` route is worth recording because it looked promising: it solved all 14
instances I threw at it and agreed with `cp.log` to 8e-7. Its canonicalization cost
rules it out, and it regresses EVaR, whose own encoding is exponential.

One, `test_cvar_on_concentrated_sample_weight_is_recovered_by_the_scs_fallback` in
`tests/test_optimization/test_convex/test_risk_budgeting.py`. Because the stall is
version-dependent, it asserts the **outcome** — the problem is solved and the weights
are a valid budget — and inspects `fallback_`/`fallback_chain_` only when the fallback
actually fired. That is what #292 is about and it holds on clarabel 0.10 and 0.11
alike. It is skipped when SCS is not installed.

The `OpinionPooling` docstring is out of `SKIPPED` in `src/conftest.py`, so the example
itself is now executed and its output verified on every run. `SKIPPED` is empty again:
every documented example runs. (A merge from `main` had silently reinstated that entry
while this PR sat open, so the example was being skipped again.)

The documented weights are pinned at three significant digits per the convention from
#283, and do not depend on which solver gets there: clarabel 0.10.0 and SCS produce
them identically to six decimals (`0.041207 0.031978 0.029986 0.044745`).

Full suite: `7763 passed`, with `ruff check` and `ruff format` clean, on clarabel
0.11.1 + scs 3.3.1.


HugoDelatte (changes requested):
Thanks for investigating the CLARABEL failure. 

I don't think the automatic retry is the optimal API as it uses a solver the user didn't choose, with parameters they cannot configure, and no way to disable it. It also takes precedence over an explicitly configured `fallback`. Someone asking to fall back to `EqualWeighted()` can now get an SCS solution instead, with `fallback_` and `fallback_chain_` both remaining `None`.

For this PR, I think it's better to use the existing `fallback` API in the example and remove the automatic solver retry. We already support retrying the same estimator with another solver:

```python
from sklearn.base import clone

model = RiskBudgeting(
    risk_measure=RiskMeasure.CVAR,
    prior_estimator=opinion_pooling,
)

model.set_params(
    fallback=clone(model).set_params(
        solver="SCS",
        solver_params={
            "eps_abs": 1e-6,
            "eps_rel": 1e-6,
            "max_iters": 100_000,
        },
    )
)
model.fit(X)
```

This preserves the prior, risk measure, constraints, and other settings, while keeping solver selection explicit. SCS gets its own absolute and relative tolerances and iteration limit. The successful fallback and the failed primary attempt are then available through `fallback_` and `fallback_chain_`.

We can consider direct solver retries in the future if we see clear benefits. Reusing the fitted prior and optimization problem could be useful for expensive or stochastic priors, or for `partial_fit`, where refitting the entire estimator is not equivalent. If those use cases justify another API, we can discuss an explicit, configurable solver sequence and how it should interact with estimator fallback. CVXPY's `solver_path` would be worth looking at then.

WDYT?

tschm:
Agreed on every point — the automatic retry is reverted in fe37741, and the PR
description is rewritten to match. Your objection that it overrides an explicitly
configured `fallback` while leaving `fallback_` and `fallback_chain_` empty is the one
I should have caught: it makes the two mechanisms compete for the same job and the
silent one wins.

`ConvexOptimization` is back to `main` verbatim — `FALLBACK_SOLVER`,
`_get_fallback_solver`, the `_solve`/`_solve_once` split and the `solver_` attribute
are all gone, as are the `solver` docstring additions in the five convex estimators.
**No library code changes in this PR any more**; the diff is the example, the doctest
configuration, the tests and the dependency groups.

The example uses your snippet, with one cosmetic difference: `model =
model.set_params(...)` rather than a bare call, only so the doctest does not have to
swallow the repr.

**One decision I made that is worth your confirmation:** since the retry is gone, I
also took `scs` out of the runtime dependencies. It now sits in the `dev` and `docs`
groups, which is how `PySCIPOpt` is already handled, and `install.rst` names SCS in
*Install Additional Solvers* alongside SCIP/GUROBI/MOSEK. That keeps the lean
`cvxpy-base` + `clarabel` install you deliberately have, and the example documents
`pip install scs` inline. Say the word if you would rather have it as a hard
dependency.

Two notes on the tests:

- The six retry-mechanism tests are replaced by a single one that asserts the
  **outcome** — the concentrated-`sample_weight` CVaR problem is solved and the weights
  are a valid budget — and only inspects `fallback_`/`fallback_chain_` when the
  fallback actually fired. On clarabel 0.10 the primary solver succeeds on its own, so
  asserting the fallback fires would fail the min-deps job.
- A merge from `main` had silently reinstated the `OpinionPooling` entry in `SKIPPED`
  while this PR sat open. It is removed again, and `SKIPPED` is now empty: every
  documented example runs.

On the forward-looking part: agreed that reusing a fitted prior and the built CVXPY
problem is the interesting case, especially for expensive or stochastic priors and for
`partial_fit`, where refitting the whole estimator is not equivalent. Happy to open a
separate issue for an explicit, configurable solver sequence and how it should compose
with estimator fallback, with `solver_path` as the starting point — this PR does not
need to prejudge it.

Local run on clarabel 0.11.1 + scs 3.3.1: `7763 passed`, `ruff check` and
`ruff format` clean.

tschm:
Opened #319 for the solver-sequence discussion, so this PR does not have to prejudge it. It covers what `solver_path` would have to settle (composition with estimator `fallback`, what counts as success given skfolio accepts `optimal_inaccurate` today, certificates short-circuiting, per-point application on frontier sweeps, mixed-integer entries) and the two caveats on reusing CVXPY's own: it accepts only `status == OPTIMAL`, and it is absent from `cvxpy-base` 1.5.0 — present from 1.6.0 — so it would mean raising the declared minimum.

HugoDelatte (dismissed):
Thanks for the update. I've made a few documentation adjustments:

- The `OpinionPooling` example now uses SCS directly, with a short explanation and a link to the fallback tutorial. This keeps the example focused on opinion pooling.
- The solver fallback demonstration is in the existing “Failure and Fallbacks” tutorial. It uses the concentrated-probability case so we can show an actual failure and recovery, followed by `fallback_` and `fallback_chain_`. There is also a note that other CLARABEL versions may solve it directly.
- I reverted the installation-page paragraph. The SCS installation instructions are next to the examples that need it.

HugoDelatte (approved):

PR #294, opened 2026-09-09
ci: run the publishing workflows only on the canonical repository

tschm:
#### Reference Issues/PRs

None. Found from a Semantic Release run that fired on the fork:
https://github.com/tschm/skfolio/actions/runs/34355840525

A fork inherits these workflows along with their triggers. `release.yml` runs on
`workflow_run` of **CI** on `main`, so a push to a fork's `main` reaches the release
job in the fork. It did, and failed at *Generate a token*:

```
Error: The 'client-id' (or deprecated 'app-id') input must be set to a non-empty
string. If using a secret or variable, ensure it is available in this workflow context.
```

That is the benign outcome, and only because the release app secrets are empty on a
fork. It is worth being precise about what the run actually attempted: it got past the
`workflow_run` conclusion check and into the release job, so with those secrets
present the same run would have created tags, published to PyPI and cut a GitHub
Release from a repository that is not the canonical one.

Three workflows publish or deploy and none was guarded:

- **`release.yml`** — tags, publishes to PyPI, creates GitHub Releases. This is the
  one that fired.
- **`docs.yml`** — deploys the documentation site to GitHub Pages. It chains from
  *Semantic Release*, so it inherits exactly the same exposure. Guarding the release
  job alone would stop it in practice today, but only as a side effect of the chain;
  the guard belongs on the job that does the publishing.
- **`export-sbom.yml`** — runs on `v*` tag pushes and calls `gh release create`, so a
  tag pushed to a fork would create a Release there. Different trigger, same class of
  problem.

Each now carries `github.repository == 'skfolio/skfolio'` on the publishing job. I
hardcoded the repository rather than testing `github.event.repository.fork`, so that
a non-fork mirror or a transferred copy does not publish either.

`ci.yml` and `dependency-review.yml` are deliberately left alone — running tests and
dependency review on a fork is the point of a fork. `ci.yml` already handles the fork
case where it matters, at the Codecov upload step.

- **Whether `docs.yml` and `export-sbom.yml` are in scope for you.** Only
  `release.yml` actually misfired; the other two are the same bug reached by a
  different path. Happy to drop them to a separate PR if you would rather keep this
  one to the observed failure.
- **The literal `'skfolio/skfolio'`** now appears in three places. If you would prefer
  it derived rather than hardcoded, say so — but note that every alternative I know of
  (`!github.event.repository.fork`, `github.repository_owner`) is weaker.

Unrelated, spotted in the same log and not touched here: `create-github-app-token@v3`
warns that `app-id` is deprecated in favour of `client-id`, which would need the
matching secret renamed, so it is yours to make rather than mine to guess at.

No — workflow conditions are not reachable from the test suite. I verified that all
five workflow files still parse and that the folded `>-` scalars collapse to a single
`${{ ... }}` expression with both conditions ANDed, rather than to two blocks:

```
docs.yml        deploy.if  = ${{ github.event.workflow_run.conclusion == 'success' && github.repository == 'skfolio/skfolio' }}
export-sbom.yml sbom.if    = ${{ github.repository == 'skfolio/skfolio' }}
release.yml     release.if = ${{ github.event.workflow_run.conclusion == 'success' && github.repository == 'skfolio/skfolio' }}
```

The `${{ }}` wrapper at job level matches what these files already do. The
`actions/runner#1173` note in `release.yml` is about *step*-level conditions on
`steps.*.outputs`, and these expressions evaluate to real booleans.


HugoDelatte (approved):
Thanks for the PR.

The three job-level repository guards are clear, appropriately scoped, and preserve existing behavior in the canonical repository.

I've shortened the release comment to clarify its intent.

PR #295, opened 2026-09-09
[ENH] Optimize Gini mean difference with constraint generation

### LINKED ISSUE #7: GINI_MEAN_DIFFERENCE Risk Measure for MeanRisk model [BUG]
GINI_MEAN_DIFFERENCE Risk Measure for MeanRisk model not working. I have tried with SCIP and CLARABEL solvers. Looks like I am missing something. Appreciate any help. Thanks!
```
model = MeanRisk(
    risk_measure=RiskMeasure.GINI_MEAN_DIFFERENCE,
    objective_function=ObjectiveFunction.MAXIMIZE_RATIO,
    portfolio_params=dict(name="Max Sharpe"), 
    solver='CLARABEL'
)
model.fit(X)
model.weights_-->
```

albystack:
## Description

Replace the quadratic OWA assignment formulation for empirical Gini mean difference (GMD) with constraint generation. Sorting candidate portfolio returns identifies the most violated permutation cut, which is reduced to asset coefficients before entering CVXPY:

$$
\theta \geq (a^T P X)w
$$

Each generated cut contains only one coefficient per asset. This avoids constructing a quadratic observation-level model on the ordinary constraint-generation path.

## Implementation

- Use a concrete `_GiniMeanDifference` helper with `solve()` and `evaluate()`, passed as one optional `gmd` through the existing solve path.
- Keep cut history local to each solve. Parameter sweeps start from the base problem; failure handling and saved-problem restoration remain in `_solve_problem`.
- Continue until normalized separation converges, without a fixed 500-cut limit. Materially violated duplicate cuts still produce a detailed solver error.
- If a relaxation is unbounded or produces a nonpositive ratio normalization factor, retry with an exact pairwise GMD epigraph. **This exceptional fallback uses quadratic memory in the number of observations**, as documented in the user guide.
- Report empirical GMD from the final normalized, possibly target-relative weights. Preserve inaccurate-solution warnings, rejection of unacceptable solver statuses, and saved primal and cone dual values after failed sweep targets.

No public API changes or new dependencies are introduced.

## Validation

- Rebased full project suite: **7,907 passed, 1 skipped** (local Python 3.13).
- Repository-wide Ruff lint and formatting checks pass.
- `git diff --check` passes.

Additional local validation compared 160 cases against an independent dense OWA formulation, covering bounded and unrestricted weights, target weights, four objectives, tied and collinear returns, and different data scales. All production and reference solves returned `optimal`; 60 cases used the exact fallback. The largest objective discrepancy after accounting for data scale was approximately `2.3e-8`.

Regression coverage includes the 500-observation reproduction, unrestricted return and utility optimization, nonpositive ratio factors, genuinely unbounded problems, parameter-sweep failures, saved-problem restoration, and CLARABEL/SCIP compatibility.

Closes #7


albystack:
Added tests for the two uncovered error paths: non-finite solver weights and overflow after GMD normalization. All 62 GMD tests pass locally.

HugoDelatte:
Thanks for the PR, I think it's the correct approach.

Could you move the target-weight normalization fix from the first commit into a separate PR, with its regression test? It fixes an existing bug independently of GMD. We can merge it first and rebase this PR.

I found two issues with the new algorithm:

- **The 500-cut limit rejects valid problems.** With `np.random.default_rng(42).normal(0.0005, 0.01, (500, 20))`, default GMD minimization fails. The original formulation succeeds. Increasing the internal limit lets the new implementation converge after 558 solves. We need a better stopping or fallback strategy.
- **The initial relaxation can be unbounded while the full problem has a finite optimum.** This happens with unrestricted weights (`min_weights=None`, `max_weights=None`) when maximizing return under a GMD cap, and when maximizing utility. The solver fails before separation can add the missing cuts.

On architecture, I’d keep this closer to `main`. I don’t think we need a generic constraint-generation abstraction for one measure:

- Keep `_GiniMeanDifference` as a concrete helper with `solve()` and `evaluate()`. Keep cut history and counters local to each solve.
- Pass one optional `gmd` through the existing solve path. `_solve` calls either `problem.solve()` or `gmd.solve()`, then extracts results and returns the final problem.
- Keep parameter sweeps, failure handling, and saved-problem handling in `_solve_problem`.
- Remove `_ConstraintGenerator`, the generator registry, `finalize_problem_values`, and `_solve_with_constraint_generation`.

Each sweep target should start from the base problem if cut history is reset. Also preserve the detailed GMD errors and the saved-variable restoration where needed.

albystack:
I've split the target-weight normalization fix and its regression test into #318. I'll rebase this PR once that is merged, and I'll be working on the two algorithm failures and simplifying the implementation over the next few days.

albystack:
Hi @HugoDelatte, I’ve addressed the two algorithm issues and simplified the implementation as you suggested. I also added regression tests and checked it against an independent formulation. The exact fallback does use quadratic memory, which I’ve documented. What do you think of this approach?

PR #305, opened 2026-09-10
feat(exceptions): add a SkfolioError base class

### LINKED ISSUE #303: Add a SkfolioError base class so callers can catch the library's errors
**Subcategory:** error handling — score 9 → 10

## Problem

`src/skfolio/exceptions.py` defines seven domain exceptions, and every one inherits directly from `Exception`:

```python
class OptimizationError(Exception): ...
class SolverError(Exception): ...
class EquationToMatrixError(Exception): ...
class GroupNotFoundError(Exception): ...
class FactorNotFoundError(Exception): ...
class DuplicateGroupsError(Exception): ...
class NonPositiveVarianceError(Exception): ...
```

There is no common base, so a caller cannot catch "any error raised by skfolio" as a class. Today the options are to enumerate all seven:

```python
from skfolio.exceptions import (
    DuplicateGroupsError, EquationToMatrixError, FactorNotFoundError,
    GroupNotFoundError, NonPositiveVarianceError, OptimizationError, SolverError,
)

try:
    model.fit(X)
except (OptimizationError, SolverError, EquationToMatrixError, ...):
    ...
```

or to catch bare `Exception` and lose the distinction from genuine bugs. For a library that is routinely wrapped in a backtest loop or a scheduled job — where the caller's intent is "skip this period and carry on" — that is a real gap.

The gap also grows with the library: every new exception silently widens the tuple that every downstream `except` clause must be updated to include.

Note that this is the only weak point in an otherwise strong error-handling design. The fallback chain in `src/skfolio/optimization/_base.py:145-160` — which captures the failure into `error_` and `fallback_chain_`, honours `raise_on_failure`, and warns with a message naming the attributes to inspect — is a good pattern and is not affected by this change.

## Suggested fix

Introduce a common base and derive the existing classes from it. This is fully backward compatible: every current `except OptimizationError` continues to work unchanged, and `except Exception` is unaffected.

```python
class SkfolioError(Exception):
    """Base class for all errors raised by skfolio."""


class OptimizationError(SkfolioError):
    """Optimization did not converge."""
```

Add `SkfolioError` to `__all__` and mention it in the API documentation as the recommended catch-all.

## done when

Every exception class in `src/skfolio/exceptions.py` derives from `SkfolioError`, `SkfolioError` is exported in `__all__`, and a test asserts that each of the seven is an instance of it.

tschm:
Closes #303.

## Problem

The seven errors in `skfolio.exceptions` each inherited `Exception` directly, so there was no way to catch "anything skfolio raises on purpose" as a class. Callers had two options, and neither is good:

```python
# enumerate all seven, and remember to update this whenever a new one is added
except (OptimizationError, SolverError, EquationToMatrixError,
        GroupNotFoundError, FactorNotFoundError, DuplicateGroupsError,
        NonPositiveVarianceError):

# or catch everything, including genuine bugs
except Exception:
```

This matters most where skfolio is wrapped in a loop whose intent is "skip this period and carry on" — a backtest, or a scheduled job — which is the same situation the fallback chain in `optimization/_base.py` already supports on the raising side, with `error_`, `fallback_chain_` and `raise_on_failure`.

The gap also grew with the library: every new exception silently widened the tuple that every downstream `except` clause had to be updated to include.

## Change

Add `SkfolioError(Exception)` as a common base and derive the seven existing errors from it. `SkfolioError` is exported in `__all__`, keeping the list's alphabetical order.

## Backward compatibility

The change is purely additive — every existing handler keeps working:

```
caught as SkfolioError: OptimizationError - did not converge
still caught as SolverError: solver failed
still caught as Exception: OptimizationError
```

All 12 existing catch sites in `src/` and `tests/` use `except <SpecificError>` and are unaffected. (`cp.SolverError` in `optimization/convex/_base.py` and `utils/stats.py` is cvxpy's own class, unrelated to skfolio's, and is untouched.)

## Tests

`tests/test_exceptions.py` — 17 tests. It discovers the exception classes from `__all__` rather than hard-coding a list, so an exception added later *without* the base fails `test_every_public_error_derives_from_base` by name, rather than silently reintroducing the enumeration problem for downstream callers. A guard test asserts the discovery finds seven classes, so a mistake in the discovery cannot quietly empty the suite.

## Verification

```
$ uvx prek run --all-files
check yaml...............................................................Passed
fix end of files.........................................................Passed
ruff-lint................................................................Passed
ruff-format..............................................................Passed

$ uv run pytest -q
7085 passed, 5 skipped in 61.06s
```

7085 vs 7068 on `main` — the 17 new tests, no regressions.

## Notes

- **No `docs/api.rst` entry.** The exceptions module is not currently documented in `api.rst` at all, so adding one means creating a new API section — a design call better left to maintainers than bundled into this fix. Happy to add it here if you would like it.
- **No CHANGELOG edit**, since `[tool.semantic_release.changelog]` generates entries from commit messages.


HugoDelatte (approved):
Thanks for the PR.

I agree, that's a good idea.

I've just shortened the docstrings, removed redundant inheritance checks, and replaced the fixed exception count with a nonempty check so future subclasses are covered automatically.

PR #306, opened 2026-09-10
docs: comment the two cycle-breaking deferred imports

### LINKED ISSUE #304: Comment the two undocumented cycle-breaking deferred imports
**Subcategory:** overall architecture — maintainability nit, no score change (currently 10)

## Context

The package's import structure is in very good shape. Analysing module-level, non-`TYPE_CHECKING` imports across all 24 subpackages finds **zero runtime import cycles**, with strictly directed layering: `typing`/`utils`/`_constants`/`exceptions` at the base, rising through `measures` → `portfolio` → `population` → `model_selection` → `optimization`. Six subpackage pairs are mutually dependent *only* through `TYPE_CHECKING`, which is the correct pattern.

That state is maintained in part by three function-local imports that break real cycles. They are load-bearing: moving any of them to module scope reintroduces a cycle.

## Problem

Only one of the three says so. `src/skfolio/utils/validation.py:296` gets it exactly right:

```python
# Imported here rather than at module scope: `skfolio.containers` imports
# `skfolio.utils.tools`, so a module-level import would close a package-level
# cycle between `skfolio.utils` and `skfolio.containers`. Annotations are
# postponed, so the `TYPE_CHECKING` import above covers the signature.
from skfolio.containers import AssetPanel, AssetPanelView
```

The other two carry no comment:

- `src/skfolio/containers/_asset_panel/_view.py:327` — `from skfolio.containers._asset_panel._panel import AssetPanel`
- `src/skfolio/model_selection/_validation.py:488` — `from skfolio.optimization._base import BaseOptimization`

To a reader — or to a future cleanup pass looking for stray non-top-level imports — these look like oversights rather than deliberate structure. Hoisting either one to module scope would reintroduce a cycle that the recent work in 085485b ("refactor(utils): break import-time package cycles") removed.

## Suggested fix

Add a comment to each, in the style already established at `utils/validation.py:296-300`, naming the specific cycle it prevents.

## done when

Both deferred imports carry a comment identifying the import cycle they break.

tschm:
Closes #304.

## Problem

The package import graph is acyclic, and three function-local imports keep it that way. They are load-bearing: hoisting any of them to module scope reintroduces a cycle.

Only one of the three said so. `src/skfolio/utils/validation.py:296` gets it right, with a comment naming the exact cycle it prevents. The other two carried no comment, so to a reader — or to a future cleanup pass looking for stray non-top-level imports — they looked like oversights rather than deliberate structure.

## Change

Add a comment to each, in the style already established at `utils/validation.py:296-300`.

## Both cycles verified, not assumed

**`containers/_asset_panel/_view.py:327`** — `_panel.py:48` imports `AssetPanelView` from `_view` at module scope. Hoisting the reverse import:

```
File "src/skfolio/containers/_asset_panel/_view.py", line 28, in <module>
    from skfolio.containers._asset_panel._panel import AssetPanel
ImportError: cannot import name 'AssetPanel' from partially initialized module
'skfolio.containers._asset_panel._panel' (most likely due to a circular import)
```

**`model_selection/_validation.py:488`** — `skfolio/optimization/__init__.py` pulls in `optimization.cluster._nco` and `optimization.ensemble._stacking`, both of which import `skfolio.model_selection` at module scope, which imports this module. Hoisting the reverse import fails the existing guard:

```
$ uv run pytest tests/test_import_cycles.py
E    cycle between: alpha, model_selection, optimization, prior, uncertainty_set
1 failed, 1 passed
```

## One thing worth flagging

The two are not equally protected, which I did not expect going in:

- The **`model_selection`** one is covered by `test_no_package_import_cycles`. Hoisting it fails that test by name, with a report pointing at the offending line.
- The **`containers`** one is **not** covered by it. That test builds its graph at subpackage granularity and skips same-subpackage edges (`if target and target != source`, `tests/test_import_cycles.py:114`), so a cycle *within* `containers` is outside its scope by construction. Hoisting the import still breaks loudly — every test fails, because the package no longer imports — but nothing explains why, which is exactly the situation the test file's docstring was written to prevent.

Extending `test_import_cycles.py` to module granularity, or adding a targeted test for this pair, would close that gap. I have left it out of this PR to keep the change to what #304 asked for; happy to add it here or as a follow-up, whichever you prefer.

## Verification

```
$ uvx prek run --all-files
check yaml...............................................................Passed
fix end of files.........................................................Passed
ruff-lint................................................................Passed
ruff-format..............................................................Passed

$ uv run pytest -q
7068 passed, 5 skipped
```

Comments only — no behaviour change.


HugoDelatte (approved):
looks good, thx

PR #308, opened 2026-09-10
[MNT] Keep doctest configuration out of wheels

### LINKED ISSUE #296: Stop shipping src/skfolio/conftest.py in the published wheel
**Subcategory:** dependency & security hygiene — score 7 → 9

## Problem

`src/skfolio/conftest.py` is included in the published wheel, and it imports `pytest` and `_pytest` at module level:

```python
import pytest
from _pytest.doctest import DoctestItem
```

`pytest` is declared in the `dev` dependency group, so an end-user install of skfolio contains a module that raises `ImportError` if anything imports it. pytest auto-discovers files named `conftest.py`, so a user running pytest in a layout where site-packages is on the collection path can hit it.

Confirmed against a locally built wheel:

```
$ uv build --wheel
$ unzip -l dist/skfolio-1.0.6-py3-none-any.whl | grep conftest
     2734  09-09-2026 13:39   skfolio/conftest.py
```

The cause is `[tool.setuptools.packages.find] where = ["src"]`, which sweeps in every `.py` under `src/skfolio`.

`deptry` reports the same thing as two DEP004 violations:

```
src/skfolio/conftest.py:11:8: DEP004 'pytest' imported but declared as a dev dependency
src/skfolio/conftest.py:13:1: DEP004 '_pytest' imported but declared as a dev dependency
```

## Suggested fix

Either exclude the module from the distribution:

```toml
[tool.setuptools.packages.find]
where = ["src"]
exclude = ["skfolio.conftest"]
```

or relocate the doctest configuration so that it is not inside the installed package, while keeping `--doctest-modules` over `src` working.

## done when

`unzip -l dist/*.whl | grep conftest` returns nothing, and `uv run pytest` still collects and passes the 1152 executed docstring examples.

russlan23:
Fixes #296.

Moves the doctest configuration from `src/skfolio/conftest.py` to `src/conftest.py`, outside the installed package. Pytest still discovers it when collecting `src`; the fixture and five existing skips are unchanged. `MANIFEST.in` explicitly retains it in the source distribution.

Adds a CI artifact check that builds both distributions, rejects any `conftest.py` in the wheel, and requires the source archive to retain the doctest configuration. No dependency or runtime API changes.

Validation in Linux x86_64 Docker with Python 3.12 and the project's uv development dependencies:

- Baseline wheel contains `skfolio/conftest.py`; the fixed wheel does not. The new CI artifact check passes.
- A clean wheel-only runtime environment imports skfolio without pytest installed.
- `uv run --no-sync pytest -q src`: 109 passed, 5 skipped.
- Full suite: 7,068 passed, 5 skipped, one solver-accuracy warning.
- Ruff check, Ruff format, and `git diff --check` pass.

An earlier baseline doctest run hit HTTP 503 downloading the public implied-volatility dataset; the subsequent doctest and full-suite runs above completed successfully. No skip or dependency override was added for that failure.

All 13 hosted checks now pass, including Python 3.10-3.14, macOS and Windows, the documentation build, and coverage.


HugoDelatte (dismissed):
Thanks for the PR!

It looks good, I've just merged main and added a small adjustment for #291: the source archive now includes both the shared root `conftest.py` and `src/conftest.py`. The distribution check verifies both files while continuing to exclude all `conftest.py` files from the wheel.

HugoDelatte (approved):

PR #309, opened 2026-09-11
[BUG] Validate optimization prediction feature order

JohnDReynolds:
#### Reference Issues/PRs

No associated issue. Open PR #286 also modifies `src/skfolio/optimization/_base.py`, but its current implementation does not address prediction feature-order validation.

This fixes a silent asset-alignment error when an optimization estimator predicts from a labeled DataFrame whose columns differ from those seen during fitting.

The common prediction path now validates fitted feature identity before applying positional weights. `InverseVolatility.fit` also establishes `n_features_in_` and `feature_names_in_`, bringing it into line with the other built-in optimizers and allowing it to work correctly as a fallback estimator.

Previously, permuting the same two labeled assets caused both `MeanRisk` and `InverseVolatility` to apply their learned weights to the wrong columns. `InverseVolatility` also failed when used as a fallback because its fitted feature metadata was missing.

No. This contribution introduces no new dependency.

Please focus on:

- whether feature identity is validated at the common prediction boundary before positional weights are applied;
- whether `InverseVolatility` establishes fitted metadata without changing its numerical behavior; and
- whether labeled, unlabeled, fallback, and failed-prediction paths retain their intended contracts.

Open PR #286 has a textual overlap in `_base.py`, but it does not implement this validation.

Yes. The regressions cover fitted metadata, correctly ordered prediction, reordered/missing/additional labeled features, rejection without fitted-state mutation, unlabeled-array compatibility, and successful `InverseVolatility` fallback behavior.

#### Regression evidence

- **Fail before:** Reordered labeled input was accepted by both `MeanRisk` and `InverseVolatility`, producing returns inconsistent with an explicitly name-aligned matrix product. `InverseVolatility` also lacked both fitted feature attributes and failed as a fallback when those attributes were copied.
- **Independent expectation:** Explicitly aligning columns to the fitted feature names reproduces the original prediction. Scikit-learn’s native `validate_data(..., reset=False)` contract rejects labeled feature mismatches before positional portfolio construction.
- **Pass after:** The regressions confirm that `InverseVolatility` records fitted metadata, same-order prediction matches the independent matrix product, reordered/missing/additional labeled inputs are rejected before construction, and rejection preserves fitted state. The fallback regression confirms that `InverseVolatility` succeeds and returns the independently expected portfolio.
- **Unchanged controls:** Correctly shaped unlabeled arrays remain positional and match the matrix-product oracle. Existing `ReturnDistribution`, population-weight, failed-portfolio, cloning, pipeline, and cross-validation behavior remains passing.

#### Verification

- Focused regressions: 6 passed in 2.56s.
- Complete optimization subsystem: 1,884 passed with 3 expected compatibility warnings in 31.93s.
- Full repository suite: 7,070 passed, 5 skipped, and the same 3 expected warnings in 69.81s.
- Ruff lint, baseline-aware Ruff formatting, `git diff --check`, and changed-line Pyright passed with no unapproved diagnostics.

#### Any other comments?

Asset labels established during fitting identify the assets associated with learned weights. Column permutation, omission, or addition must therefore not silently reinterpret those weights. Correctly shaped unlabeled arrays remain supported as positional inputs.

##### For all contributions


HugoDelatte (approved):
Thanks for the fix. I looks good.

I've just cleaned up the tests: prediction checks now cover both MeanRisk and InverseVolatility, mismatch cases are parameterized, and successful predictions must preserve fitted weights. The fallback test now checks independently calculated weights. I also removed redundant comments.

JohnDReynolds:
Thank you, Hugo, for the additional commits and stronger test coverage. I appreciate the thoughtful refinements.

PR #310, opened 2026-09-11
[BUG] Support nullable pandas returns in Portfolio

JohnDReynolds:
#### Reference Issues/PRs

No issue currently tracks this bug. PR #286 and PR #248 touch the same portfolio files but do not address nullable return conversion.

This fixes `Portfolio` construction with pandas nullable numeric return frames.

Pandas DataFrames containing nullable columns such as `Float64` can produce an object array through `np.asarray`. Passing that array to `np.isnan` raises a `TypeError`, even though an equivalent `float64` DataFrame works correctly.

The change converts an object-backed DataFrame to a floating NumPy array only when every declared column dtype is numeric. It preserves the original DataFrame, observation and asset labels, existing missing-return behavior, and the existing failure boundary for nonnumeric inputs.

No.

- Whether the conversion guard is appropriately limited to numeric pandas DataFrames.
- Preservation of asset labels, observations, and the caller-owned DataFrame.
- Equivalence between ordinary `float64`, nullable `Float64`, and mixed nullable/NumPy-backed columns.
- Preservation of the existing nonnumeric-input boundary.

Yes. Two regression tests cover a nullable `Float64` DataFrame and a mixed `Float64`/`float64` DataFrame.

#### Regression evidence

- **Fail before:** Both nullable regressions failed at `np.isnan(rets)` with `TypeError: ufunc 'isnan' not supported for the input types`.
- **Independent expectation:** Replacing only missing asset returns with zero and calculating the equal-weighted row-wise dot product produces `[0.005, 0.005, 0.02]`.
- **Pass after:** The ordinary, nullable, and mixed-column inputs all produce `[0.005, 0.005, 0.02]`. The matched regression also verifies asset labels, observations, and original-`X` identity.
- **Unchanged controls:** Ordinary NumPy-backed DataFrames and arrays, labeled weights, and nonnumeric input behavior remain unchanged.

#### Verification

- Ruff lint, baseline-aware formatting, whitespace checks, and changed-line Pyright passed.
- All 244 tests in `tests/test_portfolio/test_portfolio.py` passed.
- All 627 tests in `tests/test_portfolio` passed.
- The full repository suite passed with 7,070 tests and 5 skips.

#### Any other comments?

The conversion was also checked with the supported pandas 2.1.0 boundary and pandas 3.0.5.

##### For all contributions

##### For new estimators


HugoDelatte (approved):
Thanks for the PR. Looks good.

I've just extended it to diversification and weight drift, which also read the original `X`.

PR #311, opened 2026-09-11
[BUG] Honor absolute correlations in DropCorrelated

JohnDReynolds:
#### Reference Issues/PRs

None.

This makes `DropCorrelated(absolute=True)` apply the documented absolute-correlation semantics consistently to pair selection, ordering, and keep/drop decisions.

Previously, `self.absolute` was stored by the constructor but never used. For example, three columns with correlation matrix `[[1, -0.8, -0.5], [-0.8, 1, 0], [-0.5, 0, 1]]` and `threshold=0.75` returned support `[True, True, True]` for both `absolute=False` and `absolute=True`.

The change also validates `absolute` using the existing boolean validator.

No.

Please verify that absolute correlations are applied consistently to pair selection, ordering, and keep/drop decisions while the default signed-correlation behavior remains unchanged.

Yes. The tests cover the corrected absolute-correlation behavior, the unchanged signed-correlation behavior, and invalid `absolute` values.

#### Regression evidence

- **Fail before:** `absolute=True` retained all three assets, matching `absolute=False` even though one correlation magnitude exceeded the threshold.
- **Independent expectation:** Only the first pair has magnitude above `0.75`. Its members have absolute mean correlations of approximately `0.7667` and `0.6`, so the first asset must be removed, producing `[False, True, True]`.
- **Pass after:** `absolute=True` returns `[False, True, True]`. An invalid string value raises `ValueError` with the stable `absolute must be a boolean` contract phrase.
- **Unchanged control:** `absolute=False` continues returning `[True, True, True]` because neither raw negative correlation exceeds `0.75`.

#### Verification

- Focused regression file: 3 tests passed.
- Complete `tests/test_pre_selection` subsystem: 18 tests passed.
- Full repository suite: 7,070 passed and 5 skipped.
- Ruff lint, baseline-aware formatting, and Git whitespace checks passed.
- No new Python files or dependencies were added.

#### Any other comments?

The existing `absolute` parameter documentation already describes the intended behavior, so no public documentation change was necessary.

##### For all contributions

##### For new estimators


HugoDelatte (approved):
Thanks for the PR, looks good.

PR #313, opened 2026-09-13
[TST] Cover the untested error branches across the codebase

### LINKED ISSUE #302: Cover the untested error branches in BaseComposition and three other modules
**Subcategory:** test coverage & depth — score 9 → 10

## Problem

Overall coverage is 96% against the 95% project target in `codecov.yml`, so the bar is met. The shortfall is concentrated in error branches — the `raise` paths that no test exercises.

```
$ uv run pytest -q --cov=skfolio --cov-report=term
TOTAL   19283   712   96%
```

Lowest-covered modules, with the uncovered lines:

```
Name                                                Stmts  Miss  Cover  Missing
src/skfolio/base.py                                    70    15    79%  93, 136, 146-152, 164, 175, 183-188, 192, 195, 200
src/skfolio/attribution/_utils.py                      29     5    83%  17, 24, 37-39
src/skfolio/distribution/univariate/_selection.py      30     5    83%  61, 81-84
src/skfolio/uncertainty_set/_base.py                   37     6    84%  49, 88-92, 103
```

In `src/skfolio/base.py` the uncovered lines are almost entirely the failure paths of `BaseComposition`:

- `_validate_names` — the duplicate-name, constructor-conflict and `__`-in-name `ValueError` branches (lines 183-188, 192, 195, 200)
- `_set_params` / `_get_params` — the `TypeError`/`ValueError` fallbacks for malformed `(name, estimator)` lists (lines 146-152)

These are the branches that produce the error message a user sees when they misconfigure an ensemble estimator, so they are worth pinning: an untested error message is free to drift into something unhelpful.

## Suggested fix

Add `pytest.raises` tests asserting both the exception type and the message for each branch. These are one-liners — no fixtures or numerical setup required.

## done when

`src/skfolio/base.py`, `src/skfolio/attribution/_utils.py`, `src/skfolio/distribution/univariate/_selection.py` and `src/skfolio/uncertainty_set/_base.py` each reach at least 95% line coverage.

tschm:
Closes #302.

## What

Coverage was 96%, with the shortfall concentrated in the `raise` paths no test exercised. This adds tests for those branches and lifts the total to **99%**. The four modules named in the issue's *done when* are all at **100%**:

| Module | Before | After |
| --- | --- | --- |
| `src/skfolio/base.py` | 79% | 100% |
| `src/skfolio/attribution/_utils.py` | 83% | 100% |
| `src/skfolio/distribution/univariate/_selection.py` | 83% | 100% |
| `src/skfolio/uncertainty_set/_base.py` | 84% | 100% |

An untested error message is free to drift into something unhelpful, so each new test asserts both the exception type and the message.

## Defects found while writing the tests

Pinning these branches turned up five real bugs, fixed in this PR:

- **`BaseComposition._set_params` mutated the dict it was iterating.** It popped from `params` inside `for name in params`, raising `RuntimeError` before the replacement could complete — so replacing a named sub-estimator via `set_params` never worked. It now iterates over a snapshot.
- **`Portfolio` had no `__floor__` / `__trunc__`.** `math.floor` on a `MultiPeriodPortfolio` fell through to `np.floor`, which does not dispatch to `Portfolio`, so the constructor received ndarrays and rejected them with a `TypeError`. Both dunders are now defined, mirroring the existing `__round__`.
- **`StackingOptimization` was broken with `cv="prefit"` on numpy < 2.** The prefit path let numpy call `__array__` implicitly on the Portfolios returned by `predict`. numpy only recurses into nested objects that way from version 2 onwards, so on the floor of the supported range (numpy 1.24) it raised `ValueError: setting an array element with a sequence`. The conversion is now explicit.
- **`EWCovariance`, `EWVariance` and `EWMu` validated parameters too late.** `apply_window_size` ran before `_validate_params()`, so an invalid `window_size` surfaced the generic `"window_size must be a positive integer."` and each estimator's own specific message (`"..., got 0"`) was unreachable dead code. `_validate_params()` now runs first, which is also the correct order on its own terms — it only inspects constructor parameters.
- **`ImpliedCovariance.volatility_risk_premium_adj`** as a partial dict is rejected by `input_to_array`'s own NaN guard, not by the estimator's. Both paths are now covered, the estimator's guard via a scalar NaN.

## Unreachable branches

A handful of `raise` statements cannot be reached: `match` fallbacks where every enum member is already handled, and guards an earlier validator has already enforced. These are marked `# pragma: no cover` with the reason inline, rather than papered over with a test that cannot fail.

## Verification

The suite was run against two environments, since the first push failed CI on assumptions that only held in a drifted local venv:

- **Current deps** (venv synced to `uv.lock`, so the installed solvers match CI): 7568 passed, 5 skipped, coverage 99%.
- **Minimum deps** (`--resolution lowest-direct`, numpy 1.24.0 / plotly 6.0.0 / Python 3.10, matching the `min` CI job): 7459 passed.

Tests now avoid environment-specific assumptions: they use SCIPY rather than SCS (skfolio depends on `cvxpy-base`, which bundles no solvers), assert that a singular design's condition number is flagged rather than finite (LAPACK reports either a huge value or `inf` depending on the BLAS build), and assert on the set of plotly trace names rather than on trace order.

`ruff check`, `ruff format --check` and `pre-commit run --all-files` are clean.


tschm:
@HugoDelatte That's a big one but the findings are interesting. Not sure whether you really want to support numpy 1.x

HugoDelatte (approved):
Thanks for the PR!

It looks good.

I've just added some cleanups:

- The Schur repair test previously passed with NaN weights. The repair logic checked the left block twice and stored blocks before repairing them. Both blocks are now checked, and repaired matrices are saved for subsequent recursive splits.
- Schur tests now verify finite, normalized weights, repair propagation, and clean rejection of unrepairable blocks.
- Multiperiod rounding tests check actual weights, including negative fractions.
- Removed the incorrect “unreachable” exclusion in MeanRisk and added a test for changing the objective between incremental fits.
- Removed tests that merely execute empty abstract methods, replaced those bodies with ellipsis, and kept coverage exclusions narrowly scoped.
- Removed redundant guards and replaced validation-bypassing mocks with ordinary invalid-input tests.

HugoDelatte:
@tschm on numpy 1.x, we should indeed retire it. Scientific Python’s recommended dependency support window is two years and NumPy 1.x passed that window in September 2025.
We will do it in a dedicated PR.

PR #314, opened 2026-09-14
docs(attribution,population): make skipped doctest examples runnable

### LINKED ISSUE #298: Make the five skipped docstring examples runnable
**Subcategory:** executable documentation — score 8 → 10

## Problem

`pyproject.toml` runs `--doctest-modules` over `src`, so 1152 of the 1157 docstring examples execute and have their documented output verified on every CI run. That is excellent coverage.

The remaining five are skipped in `src/skfolio/conftest.py`. Each skip carries a documented reason, which is good discipline — but a skipped example is still a published example that a reader cannot run.

One of the five is a genuine failure rather than a fragment:

```python
"skfolio.prior._opinion_pooling.OpinionPooling": (
    "example raises: the pooling itself fits, but `RiskBudgeting(CVaR)` on the "
    "pooled distribution hits `SolverError: Solver 'CLARABEL' failed`"
),
```

The other four are illustrative fragments that reference names never defined in the example:

- `skfolio.attribution._predicted.predicted_factor_attribution` — `factor_returns` is never defined
- `skfolio.attribution._realized.realized_factor_attribution` — `factor_returns` is never defined
- `skfolio.attribution._realized.rolling_realized_factor_attribution` — `factor_returns` is never defined
- `skfolio.population._population.Population.boxplot_measure` — `population` is never defined

All five appear in public API documentation, where a reader's reasonable expectation is that the example runs as written.

## Suggested fix

For the four fragments, add the missing setup lines so each example is self-contained and copy-pasteable. For `OpinionPooling`, either choose a risk measure or solver configuration that converges on the pooled distribution, or adjust the example so it does not depend on the failing solve.

## done when

`SKIPPED` in `src/skfolio/conftest.py` is empty and `uv run pytest` passes with all 1157 examples executed.

anmol0705:
## Problem

Five docstring examples are currently skipped from the `--doctest-modules` CI run via the `SKIPPED` dict in `src/conftest.py`. Four of them are skipped because they reference variables that are never defined in the example (illustrative fragments), making them uncopy-pasteable despite appearing in the public API docs. This PR fixes those four; the fifth (`OpinionPooling`, a genuine solver bug) is left untouched and out of scope.

## Changes

- `predicted_factor_attribution`: defined `loading_matrix`, `factor_cov`, `idio_cov`, `asset_names`, and added the missing required `asset_names` argument to all three example calls. Also corrected `factor_mu` to per-period (daily) magnitude to match the function's documented periodicity contract.
- `realized_factor_attribution` / `rolling_realized_factor_attribution`: defined `factor_returns`, `portfolio_returns`, `loading_matrix`, `weights`, `residuals`, `daily_weights`, `asset_names` using `np.random.default_rng` (matching existing house style), and added the missing required `asset_names` argument.
- `Population.boxplot_measure`: built a real `Population` from two fitted, tagged portfolios (`InverseVolatility`, `RiskBudgeting`) following the same pattern used elsewhere in the codebase's examples, added missing imports, and fixed a call to `plot_measure_box`, which does not exist on `Population` (only `boxplot_measure` does).
- Removed the four corresponding entries from `SKIPPED` in `src/conftest.py`.

## Testing

```
uv run pytest --doctest-modules src/skfolio/attribution/_predicted.py src/skfolio/attribution/_realized.py src/skfolio/population/_population.py -v
```
All 4 previously-skipped doctests pass.

```
uv run pytest tests/test_attribution tests/test_population -q
```
317 passed, no regressions.

`uv run ruff check` / `uv run ruff format` clean on all changed files.

Confirmed (via `git stash` comparison against `main`) that one pre-existing, unrelated doctest failure in `src/skfolio/distribution/univariate/_normal_inverse_gaussian.py` is not introduced by this change.

## Out of scope

The fifth skipped example (`skfolio.prior._opinion_pooling.OpinionPooling`) is a genuine solver failure (`RiskBudgeting(CVaR)` on a pooled distribution raises `SolverError`), not a documentation issue, and is left skipped as-is.

Fixes #298


HugoDelatte (approved):
Thanks for the PR, it looks good.

I've just simplified the examples a little bit.

PR #315, opened 2026-09-14
[BUG] Validate portfolio constructor sample weights

JohnDReynolds:
#### Reference Issues/PRs

No semantic issue or pull request exists. PR [#286](https://github.com/skfolio/skfolio/pull/286) is merged in accepted main and retained the bypass. Open PR [#248](https://github.com/skfolio/skfolio/pull/248) at `c055f702` adds an independent `asset_groups` constructor field and treemap behavior; it creates textual but not semantic overlap. Open `sample_weight` search matches #293 and #255, which touch optimization and measures respectively, not portfolio construction.

Reuse the existing `sample_weight` setter contract during every public `BasePortfolio` construction path, while validating `MultiPeriodPortfolio` only after its concatenated observations are installed.

`BasePortfolio`, `Portfolio`, `FailedPortfolio`, and `MultiPeriodPortfolio` all accept wrong-length, two-dimensional, and wrong-sum constructor weights that the identical public property assignment rejects. A three-observation portfolio with all-one weights silently reports mean `0.058` instead of the independently weighted `0.022`; a two-dimensional weight returns an array instead of a scalar, and a wrong-length weight degrades the measure to NaN with a warning.

No new dependency.

Please concentrate on whether validation occurs at the correct observation-aware point for both shared and multi-period construction, and whether the representative public paths plus dimension, observation-count, and sum regressions cover the constructor contract.

Yes. See the deterministic cases and outcomes under Regression evidence.

#### Regression evidence

- **Fail before:** Every public constructor path accepts all three invalid classes. Wrong-length weighted mean becomes NaN, wrong-dimensional weighted mean becomes a one-element array, and wrong-sum weights silently scale the result.
- **Independent expectation:** The existing setter rejects the identical values with stable dimension, observation-count, or sum errors. For returns `[0.018, 0.008, 0.032]` and weights `[0.2, 0.3, 0.5]`, a direct dot product gives the accepted weighted mean `0.022`.
- **Pass after:** The same 10 constructor regressions now pass. All public constructor paths reject wrong-length, two-dimensional, and wrong-sum weights through the existing setter; valid normalized weights retain the direct-dot-product mean `0.022`.
- **Unchanged controls:** Valid normalized arrays, `None`, setter rejection, ordinary unweighted measures, valid pickling/reconstruction, and `Population.set_portfolio_params` remain controls.

#### Any other comments?

Constructor timing cannot bypass the documented one-dimensional, observation-aligned, sum-to-one contract. Valid normalized weights and `None` remain accepted; weighted measures retain their scalar type and independently derived value.

##### For all contributions


HugoDelatte (approved):
Thanks for the fix. I agree with it.

I've just added a follow-up for an integration issue: CharacteristicsFactorModel can truncate scenario probabilities through max_history without renormalizing them. With EntropyPooling, the new constructor validation then rejects a previously working hierarchical optimization. The retained probabilities are now normalized, with coverage for integer weights, individual zero weights, small positive totals, and zero retained mass.

I also added list-input variance coverage, simplified the FailedPortfolio test, corrected the getter annotation, and clarified the deferred-validation comment.

PR #316, opened 2026-09-14
docs: make docstring examples reproduce with NumPy's default print options

### LINKED ISSUE #307: Docstring examples do not reproduce outside the test suite's print options
[rhiza](https://github.com/Jebel-Quant/rhiza-claude)

## Problem

`src/skfolio/conftest.py` sets global NumPy print options for the duration of every doctest:

```python
np.printoptions(
    precision=4 if compact_arrays else 8,
    suppress=compact_arrays or ...,
    legacy="1.25",
)
```

The examples therefore pass in CI while showing output a reader cannot reproduce. Someone who copy-pastes an example into a fresh session has NumPy's defaults, and sees something different from what the docstring promises.

```
docstring:  [0.0021 0.     0.     0.1559 0.0641]
reader:     [2.07410128e-03 1.80791598e-12 1.17330564e-11 1.55908109e-01
             6.40784682e-02]
```

This is the failure mode with the longest half-life in documentation: it renders perfectly, CI is green, and the person who discovers it is a newcomer at the worst possible moment.

## Scope: 9 of 114 examples

Measured by removing only the `np.printoptions(...)` context from the fixture — keeping `chdir` and the sklearn config — and running the doctests:

```
$ uv run pytest -q src
9 failed, 100 passed, 5 skipped
```

The nine:

```
model_selection/_online/_search.py       OnlineRandomizedSearch
optimization/cluster/_nco.py             NestedClustersOptimization
optimization/cluster/hierarchical/_schur.py  SchurComplementary
optimization/convex/_distributionally_robust.py  DistributionallyRobustCVaR
optimization/convex/_maximum_diversification.py  MaximumDiversification
optimization/convex/_mean_risk.py        MeanRisk
optimization/convex/_risk_budgeting.py   RiskBudgeting
prior/_entropy_pooling.py                EntropyPooling
prior/_synthetic_data.py                 SyntheticData
```

These are the library's most prominent entry points — `MeanRisk` above all — so the examples that do not reproduce are the ones a new user meets first.

## Not all of the fixture is the problem

Two parts are legitimate and invisible to readers, and should stay:

- `sklearn.config_context(transform_output="default")` restores sklearn's *own* default, undoing leakage from unit tests that call `set_config` without restoring it. A reader in a fresh interpreter already has this; removing it would make the doctests order-dependent.
- `monkeypatch.chdir(tmp_path)` keeps `AssetPanel.save("asset_panel")` out of the working tree. It changes no printed output.

Only the print options substitute a different environment for the reader's.

## Proposed fix for eight of the nine

Move the formatting into the example, where the reader can see it. This produces **byte-identical** output with no global state:

```python
>>> print(model.weights_.round(4))
[0.0021 0.     0.     0.1559 0.0641]
```

`.round(4)` also removes the need for `suppress=True`: a value of `1.8e-12` rounded to four places is genuinely `0.`, so scientific notation never arises.

There is a reasonable objection that this is "code written only to satisfy doctest", which `CONTRIBUTING.md` asks contributors to avoid. It reads the other way round: `.round(4)` is what a user would actually type to inspect weights, whereas a hidden global print option is machinery existing purely to satisfy doctest — with the added cost that the reader cannot see it.

## The ninth case is genuinely harder

`OnlineRandomizedSearch` is the only example that depends on `legacy="1.25"`, and it is not a formatting preference. It prints a dict of NumPy scalars:

```python
>>> search.best_params_
{'prior_estimator__covariance_estimator__half_life': 63.31...,
 'prior_estimator__mu_estimator__half_life': 14.18...}
```

NumPy 2.0 changed scalar repr, so on NumPy 2.x a reader sees `np.float64(63.31...)` instead. `pyproject.toml` requires `numpy>=1.23.5`, spanning the change, so **no single literal output is correct for every supported version** — which is presumably why `legacy="1.25"` is set, and the conftest docstring says so explicitly ("NumPy scalars display as plain numbers, including inside dictionaries").

That makes this a real constraint rather than an oversight. The version-independent option is to avoid reprs of NumPy scalars in examples, e.g.:

```python
>>> {k: float(v) for k, v in sorted(search.best_params_.items())}
```

which prints plain floats on every supported NumPy. Whether that is worth the extra noise in this one example is a maintainer call.

## Suggested sequencing

The eight array cases are mechanical and independently reviewable. The `legacy` case can be decided separately, and `legacy="1.25"` kept until then — it is load-bearing for exactly one example, verified by removing it alone:

```
$ uv run pytest -q src        # print options intact, legacy removed
1 failed, 108 passed, 5 skipped
```

## done when

The eight array examples show their own formatting and pass with the `precision`/`suppress` options removed from `src/skfolio/conftest.py`, and a decision is recorded on the `OnlineRandomizedSearch` / `legacy="1.25"` case.

## Related

- #296 is independent: the conftest has to live in the doctest tree either way, and moving it out of the installed package does not change what it configures.
- #298 covers the five examples skipped outright, which is the same concern (documentation that does not run) from the other direction.

wolfgang-aura:
docs: make docstring examples reproduce with NumPy's default print options

Closes #307.

Nine docstring examples pass in CI while showing output a reader cannot reproduce. `src/conftest.py` wrapped every doctest in `np.printoptions(precision=4|8, suppress=..., legacy="1.25")`, so `print(model.weights_)` documented `[0.0021 0. 0. 0.1559 0.0641]` while a fresh interpreter prints `[2.07410128e-03 1.80791598e-12 ...]`. The examples that do not reproduce are `MeanRisk`, `RiskBudgeting`, `MaximumDiversification`, `DistributionallyRobustCVaR`, `NestedClustersOptimization`, `SchurComplementary`, `EntropyPooling`, `SyntheticData` and `OnlineRandomizedSearch`.

The fixture now pins NumPy's own defaults (`precision=8, suppress=False, legacy=False`) instead of a private format. It stays a context rather than being removed because `tests/conftest.py` sets `np.set_printoptions(suppress=True, precision=6)` for the whole session; CI runs `tests` and `src` together, and without the pin that leak lands on the doctests. The nine examples now round in the example itself (`print(model.weights_.round(4))`), which is what a reader would type, and `OnlineRandomizedSearch.best_params_` prints plain floats instead of relying on `legacy="1.25"` to hide `np.float64(...)`. One `EntropyPooling` expected output is split where NumPy wraps it at the default line width. Ten files, +27/-34, no library code changed.

### How this was tested

- Reproducer: `doctest.testmod` over the nine modules with NumPy defaults and no conftest, at `5ee2913a236b`: 15 failed examples across 9 modules. With this change: 0 failed of 197.
- Gate: `pytest -q -p no:cacheprovider src/conftest.py` plus the nine modules: 10 passed before and after.
- Full doctest tree, `pytest -q -p no:cacheprovider src`: 109 passed, 5 skipped.

All results are from one host: Windows 11, CPython 3.14.3, NumPy 2.5.3. CI on Linux and macOS is the cross-platform check.

### Disclosure

> This change was drafted with AI assistance (claude (`claude-opus-5`) wrote the patch, claude (`claude-opus-5`) reviewed it), running under Mailman. The test results above come from the harness executing the commands itself, not from either agent's account of its own work. The account filing this pull request takes responsibility for every line of it.


HugoDelatte:
That was done on purpose, it's a trade-off: we favour docstring readability versus matching exactly what a user would see in a default copy/paste action (which is not really what the docstrings are for anyway).

wolfgang-aura:
Understood, thanks for the quick look.

HugoDelatte:
@wolfgang-aura thanks to you for opening the PR. Actually, I can see that your agent opened it based on an issue he encountered, are you able to share more details about it (maybe this doctstring formating creates issues I'm unaware of).

wolfgang-aura:
No real-world problem on my side — the PR was filed against #307, which someone else reported, and the only thing the run added was a machine check of that report: with NumPy's default print options, 15 examples across the nine modules named there fail, 0 after the change. Nothing beyond what #307 already describes, so if the trade-off is deliberate there's nothing further from me. Thanks again.

PR #317, opened 2026-09-14
[BUG] Preserve MultiPeriodPortfolio arithmetic state

JohnDReynolds:
#### Reference Issues/PRs

No issue currently claims this arithmetic-state defect.

PR #313 modifies the same files but addresses the separate `floor` and `trunc` failures. PR #315 modifies `MultiPeriodPortfolio` construction to validate `sample_weight`. Neither supplies this fix.

Arithmetic on a configured `MultiPeriodPortfolio` transforms its child portfolios correctly but resets 15 top-level configuration fields, including its name, annualization factor, compounding convention, sample weights, risk-free rate, risk parameters, and observation-order policy.

This change:

- reconstructs arithmetic results using the existing constructor-state mapping while replacing only the transformed child portfolios;
- preserves the left operand’s complete configuration and identity;
- makes addition and subtraction reject operands with conflicting non-identity configuration; and
- retains the existing rule that `name` and `tag` may differ, with the left operand’s values winning.

No new dependency is introduced.

Please verify that every successful arithmetic reconstruction preserves the left operand’s complete constructor state and that binary compatibility matches `Portfolio`.

The separate `floor`/`trunc` repair from PR #313 and constructor validation from PR #315 remain outside this PR.

Yes. The regression coverage includes 12 arithmetic forms, left-operand identity behavior, all 30 addition/subtraction configuration-conflict combinations, transformed child values, and operand preservation after rejected operations.

#### Regression evidence

- **Fail before:** Twelve successful arithmetic forms produced the correct child values but reset 15 configured fields. Addition and subtraction also accepted operands with conflicting measurement, weighting, risk, and validation configuration.
- **Independent expectation:** Child results are compared with direct `Portfolio` arithmetic, while top-level state is compared with the original constructor mapping and the established `Portfolio` compatibility contract.
- **Pass after:** All 12 forms preserve every constructor field except the intentionally transformed child collection. Addition and subtraction preserve left identity metadata and reject all 30 incompatible configurations without mutating either operand.
- **Verification:** Ruff, formatting, whitespace, and changed-line Pyright passed. The focused/downstream checks passed 284 tests, the portfolio subsystem passed 737 tests, and the full suite passed 7,297 tests with 5 skipped.

#### Any other comments?

This PR does not change public signatures, return-value mathematics, dependencies, exports, documentation, collection mutation, or sample-weight validation.

##### For all contributions

##### For new estimators

Not applicable; this PR adds no estimator.


HugoDelatte (approved):
Thanks for the PR. I agree with it and it looks good.

I made a few adjustments:
- Allow different `check_observations_order` values and preserve the left operand’s preference. This controls validation rather than how performance is measured.
- Rename the construction helper to `_create_from_child_portfolios`.
- Parameterize the arithmetic tests using small synthetic portfolios, with an explicit regression test for monthly annualization.

JohnDReynolds:
Thank you for the review and the thoughtful follow-up commits, Hugo. I especially appreciate the refined check_observations_order semantics and the clearer, more focused regression coverage.

PR #318, opened 2026-09-15
[BUG] Normalize target weights in maximum-ratio optimization

albystack:
#### Reference Issues/PRs

Extracted from #295 following maintainer feedback, so this independent bug fix can be merged first.

With `MAXIMIZE_RATIO` and `target_weights`, risk was computed using `w - target_weights` even though `w` is in homogeneous coordinates. After normalization, this scales the target allocation incorrectly.

Use `w - target_weights * factor` so normalized risk is computed from `weights_ - target_weights`. Objectives whose factor is 1 retain the same expression.

No.

The target-weight scaling under the maximum-ratio transformation and the regression test's independent calculation of active-return standard deviation.

Yes. The regression test compares reported standard-deviation risk with the sample standard deviation of returns computed from the normalized active weights.

Validation on this branch:

- Target-weight tests: 53 passed.
- Ruff lint and formatting checks passed for both changed files.
- `git diff --check` passed.

#### Any other comments?

This contains only the target-weight fix and its regression test from the first commit of #295. The GMD PR will be rebased after this fix is merged.

##### For all contributions

##### For new estimators

Not applicable.


HugoDelatte (approved):
Thanks for the PR!

The normalization fix is correct.

I expanded the regression test to cover standard deviation, variance, and CVaR, comparing the reported risk with a separate calculation from the final active weights. This covers both linear and squared risk normalization, plus the CVaR formulation. I also shortened the explanatory comment.

PR #320, opened 2026-09-16
build: document the indirect runtime dependencies and gate deptry in CI

### LINKED ISSUE #297: Document clarabel and joblib as indirect runtime dependencies and gate deptry in CI
**Subcategory:** dependency & security hygiene — score 7 → 9

## Problem

`deptry` reports two unused runtime dependencies:

```
pyproject.toml: DEP002 'clarabel' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'joblib'   defined as a dependency but not used in the codebase
```

Both are genuine runtime requirements reached indirectly, so the declarations are correct and should stay:

- `clarabel` is reached through cvxpy's solver-name dispatch — it is never imported, only named as a string (`solver: str = "CLARABEL"`, `src/skfolio/optimization/convex/_base.py:466`).
- `joblib` is reached through `sklearn.utils.parallel`; `skfolio` itself never imports it. Every reference in `src/` is a docstring or comment.

Nothing in the repository records either path, so the tool cannot tell a deliberate indirect dependency from a stale one — and no CI job runs `deptry` at all, so the dependency surface is currently ungated.

## Suggested fix

Add a `[tool.deptry]` section with per-rule ignores and a comment naming the indirect path for each, then invoke it from CI alongside the existing lint job.

## done when

`uv run deptry src` exits 0 with the two indirect dependencies explicitly ignored and commented, and a CI step runs it on every pull request.

tschm:
#### Reference Issues/PRs

Fixes #297.

`deptry` reported two unused runtime dependencies:

```
pyproject.toml: DEP002 'clarabel' defined as a dependency but not used in the codebase
pyproject.toml: DEP002 'joblib'   defined as a dependency but not used in the codebase
```

Both are genuine runtime requirements that are never imported, which I verified rather
than assumed:

- **`clarabel`** — reached through cvxpy's solver-name dispatch. `grep -rn "import clarabel" src` finds nothing; it appears only as the string `"CLARABEL"` (`solver: str = "CLARABEL"` and the per-solver parameter defaults).
- **`joblib`** — reached through `sklearn.utils.parallel`, which skfolio calls for `n_jobs` (`import sklearn.utils.parallel as skp` … `skp.Parallel(n_jobs=self.n_jobs)`). Every `joblib` occurrence in `src` is a docstring or a comment.

So the declarations are correct and stay; what was missing is the record of *why*.
They now sit in `[tool.deptry.per_rule_ignores]` with the indirect path named in a
comment.

**One finding beyond the issue.** A clean run also reported two `DEP004`s that postdate
the issue, from `src/conftest.py` (added in #308):

```
src/conftest.py:13:8: DEP004 'pytest' imported but declared as a dev dependency
src/conftest.py:15:1: DEP004 '_pytest' imported but declared as a dev dependency
```

That file is doctest configuration deliberately kept out of the wheel — the existing
"Check distribution contents" step asserts exactly that — so it is development tooling
that happens to live under `src`, not package code. It is excluded from the scan rather
than ignored per-rule, since the issue is the file's classification and not the imports.

#### Gating

`deptry` joins the `dev` group and the lint job runs it on every pull request.
Two deliberate choices:

- **It runs in the project environment**, not as a standalone `uvx` run like Ruff.
  deptry maps top-level modules to their distributions through installed metadata
  (`cvxpy` comes from `cvxpy-base`), so without the environment it reports false
  positives.
- **It runs after the distribution-contents step**, so the `uv build` there still sees
  a tree with no `.venv` in it.

`CONTRIBUTING.md` documents `uv run deptry src` and where an indirect dependency
belongs.

No unit tests — the change is configuration and a CI step. Verified directly:
`uv run deptry src` exits 0 on this branch, and reintroducing an unused runtime
dependency (`rich`) makes it report `DEP002` and exit 1, so the gate is not vacuous.
`ruff check` and `ruff format` clean.


HugoDelatte (approved):
I kept deptry as a development dependency and made a few adjustments:

- Moved the check into the existing Linux job with the main Python version and latest dependencies, avoiding another environment installation.
- Removed the CONTRIBUTING addition to keep the guide concise. Dependency changes still need individual review.
- Removed joblib from our direct dependencies. Skfolio now uses scikit-learn’s parallel API, and scikit-learn already requires joblib. Updated the installation guide and removed the deptry exception.
- Kept the clarabel exception: it is our default solver, loaded by CVXPY without a direct import.

PR #321, opened 2026-09-16
feat(optimization): add an explicit solver sequence to the convex estimators

### LINKED ISSUE #319: [ENH] Explicit, configurable solver sequence for convex optimizations
Follow-up from the review discussion in #293 (and #292). Opening it so the PR does not
have to prejudge the design.

**Is your feature request related to a problem? Please describe.**

When a solve fails, the recovery path today is the estimator-level `fallback`: it
refits the whole estimator with a different `solver`. That is the right answer for the
documented `OpinionPooling` example and is what #293 ships. There are three cases where
it is either wasteful or not equivalent:

1. **Expensive priors.** `EntropyPooling` solves its own convex problem, and
   `OpinionPooling` fits one per opinion. A fallback repeats all of that just to change
   one argument to `problem.solve`.
2. **Stochastic priors.** With `SyntheticData` or a bootstrapped prior, the refit fits a
   *different* distribution than the one that failed, so the recovered weights do not
   correspond to the problem that could not be solved.
3. **`partial_fit`.** `MeanRisk.partial_fit` exists, but the fallback wrapper only wraps
   `fit` (`cls.__dict__.get("fit")` in `BaseOptimization.__init_subclass__`, and
   `_run_fallback_chain` calls `fb_est.fit`). So a failed solve during a streaming
   update has **no** recovery path at all, and a full refit is not equivalent to an
   incremental one.

In all three, everything upstream of the solve is already computed and still valid;
the only thing that needs to change is which solver finishes the problem.

**Describe the solution you'd like**

An explicit, opt-in, ordered solver sequence on `ConvexOptimization`, each entry
carrying its own parameters:

```python
MeanRisk(
    solver_path=[
        ("CLARABEL", {"tol_gap_abs": 1e-9}),
        ("SCS", {"eps_abs": 1e-6, "eps_rel": 1e-6, "max_iters": 100_000}),
    ]
)
```

mutually exclusive with `solver` / `solver_params`, plus a fitted attribute reporting
which entry produced the solution. The semantics worth settling before any code:

- **Composition with `fallback`.** The sequence is the inner loop and estimator
  `fallback` stays the outer one: exhaust the sequence, then run the fallback chain.
  `fallback_chain_` should then record that the whole sequence failed, not just its last
  entry.
- **What counts as success.** skfolio currently accepts a non-`OPTIMAL` status and warns
  `"Solution may be inaccurate"`. If the sequence advanced on `optimal_inaccurate` it
  would change behaviour for problems that succeed today, so the conservative rule is to
  advance only on `SolverError` and keep the existing acceptance.
- **Certificates short-circuit.** An infeasible or unbounded result is a property of the
  problem, not of the solver, so it should abandon the rest of the sequence rather than
  retry every entry to reach the same verdict.
- **The multi-optimization path.** For parameter sweeps (`n_optimizations > 1` in
  `_solve_problem`, e.g. an efficient frontier) the sequence should apply per point, so
  one hard point does not sink the whole sweep.
- **Mixed-integer problems.** Entries that cannot handle integer variables need to be
  skipped or rejected up front rather than failing at solve time.

**Describe alternatives you've considered**

- **CVXPY's own `problem.solve(solver_path=[...])`** — the closest fit: it takes
  `list[str | tuple[str, dict]]`, tries them in order and returns the first success.
  Two caveats. It accepts only `status == OPTIMAL`, so results skfolio accepts today
  with the inaccuracy warning would be discarded; and it is not in `cvxpy-base` 1.5.0
  (verified absent from 1.5.0's `problems/problem.py`, present from 1.6.0), so using it
  means raising the declared minimum from `cvxpy-base>=1.5.0` to `>=1.6.0` or guarding
  by version. It is also mutually exclusive with `solver=` there, which the API above
  would have to mirror anyway.
- **An automatic, unconfigurable retry** — implemented in the first revision of #293 and
  rejected in review: it picks a solver the user did not name, cannot be disabled, and
  silently takes precedence over an explicitly configured `fallback`.
- **Status quo** — estimator `fallback` only. Sufficient for the documented example;
  this issue is only about the three cases above.

**Additional context**

The trigger was #292: CVaR risk budgeting on a `sample_weight` concentrated by
`EntropyPooling` views. clarabel 0.10.0 solves it; 0.11 stalls and terminates in
`InsufficientProgress` with the dual residual frozen at `4.15e-02` and a step size of
`0.00e+00`. SCS solves it, and across the 48 risk-measure × prior combinations tried in
#293 the two solvers agree wherever both succeed, with no combination failing both.

Note that `scs` is a `dev`/`docs` dependency only, so any solver sequence shipping SCS
as a default entry would have to reopen that decision.

tschm:
#### Reference Issues/PRs

Fixes #319. Independent of #293 — that PR keeps the `OpinionPooling` example on the
estimator `fallback`, as requested in review, and this one does not touch it. They do
not conflict, and this can wait behind it or land on its own.

A convex problem can be well posed and still defeat one algorithm. CVaR risk budgeting
on a `sample_weight` concentrated by `EntropyPooling` views (#292) is bounded and
feasible, yet CLARABEL 0.11 freezes with a non-zero dual residual and terminates in
`InsufficientProgress`, while the first-order method SCS solves it. The reverse also
happens, so neither is a better default.

Recovering such a solve today means the estimator-level `fallback`, which refits
everything. Per #319 that is the wrong tool in three cases: an expensive prior
(`EntropyPooling` solves its own convex problem) repeats all of it to change one
argument; a stochastic prior refits a *different* distribution than the one that
failed, so the recovered weights do not answer the failed problem; and `partial_fit`
has no fallback at all, since the wrapper only wraps `fit`. In each, everything
upstream of the solve is already computed and valid.

So convex estimators accept `solver_path`, an ordered list of solvers used in place of
`solver`:

```python
model = RiskBudgeting(
    risk_measure=RiskMeasure.CVAR,
    solver_path=[
        "CLARABEL",
        ("SCS", {"eps_abs": 1e-6, "eps_rel": 1e-6, "max_iters": 100_000}),
    ],
)
model.fit(X)
model.solver_   # the solver that produced the solution
```

An element is either a solver name, which takes the same default parameters as `solver`
would, or a `(solver, solver_params)` tuple carrying its own.

#### Semantics

These are the points #319 listed as needing a decision. Each is implemented as
described there; say the word on any of them and I will change it.

- **Nothing changes by default.** A single-solver path — which is what every existing
  estimator has — behaves exactly like `solver` does today: the same error, no warning,
  no extra solve. `solver_` is then simply `solver`.
- **Only failure advances the sequence**, and every advance warns, so a solution is
  never silently produced by a solver the user did not list. A solution CVXPY returns
  as inaccurate is accepted exactly as it is today, with the existing
  `"Solution may be inaccurate"` warning; advancing on it would change results for
  problems that succeed today.
- **A certificate stops the sequence.** Infeasible or unbounded is a property of the
  problem, not of the solver, so the remaining entries are not tried and the error stays
  anchored on the solver that proved it.
- **Mixed-integer problems skip the entries that cannot express integer variables**,
  with a warning naming them; when none remains, the existing
  `"require a mixed-integer solver"` error stands unchanged. Skipping rather than
  erroring is the reading I chose for an ordered preference list — it declines a solver
  the user named, it never introduces one they did not.
- **It applies per optimization**, so on an efficient frontier one hard point does not
  sink the sweep, and `solver_` is then the list of solvers aligned with the
  optimizations (`None` where a point failed and `raise_on_failure=False`).
- **It cannot be combined with a non-default `solver` or with `solver_params`**, which
  would make the first attempt ambiguous. `solver` has a default, so only a deviation
  from `"CLARABEL"` counts as a conflict.

#### On CVXPY's own `solver_path`

Not used, for the two reasons in #319, both verified rather than assumed: it accepts
only `status == OPTIMAL`, so the inaccurate solutions skfolio accepts today would be
discarded; and it is absent from `cvxpy-base` 1.5.0's `problems/problem.py` and present
from 1.6.0, so it would mean raising the declared minimum. The parameter keeps CVXPY's
name and element format so the two read the same.

#### Refactor included

The per-solver default parameters moved out of `match self.solver` blocks in three
estimators into a `_default_solver_params` class mapping (`MeanRisk` overrides it to add
its SCIP tuning). That is what lets a path entry be tuned exactly like the equivalent
`solver`, instead of only the primary one. `_solver_params` keeps its meaning as the
parameters of the first attempt, so the tests asserting it are untouched.

26, in `tests/test_optimization/test_convex/test_solver_path.py`. They avoid depending
on any solver's numerical behaviour: the failing solver is `SCIPY`, which ships with
cvxpy and genuinely cannot take the exponential cone risk budgeting needs, so the
failure is real and deterministic on every platform and every dependency set — no
monkeypatching and no version-sensitive stall.

Covered: the default single attempt; advancing on failure; the first success ending the
sequence; string entries taking their solver's defaults; tuple entries carrying their
own; the exhausted-path error naming every attempt; a single-entry path raising the
plain error with no warning; the certificate short-circuit; per-optimization
application on a frontier; mixed-integer skipping and its no-solver-left error;
uninstalled entries; the `solver`/`solver_params` conflicts; and malformed paths. One
test drives the real #292 instance through `["CLARABEL", ("SCS", …)]`, asserting the
outcome rather than the advance, since CLARABEL 0.10 solves it alone; it skips when SCS
is not installed.

Full suite: `7852 passed, 1 skipped`, `ruff check` and `ruff format` clean.


dcelisgarza:
I took this approach in my own library and it works quite well. In fact, i did this before realising I could implement an estimator fallback as well. Having both is quite nice, I'd encourage the maintainers to consider it. I've found it especially useful for pareto fronts, MIP problems, or tricky risk measures as the PR states. 

There's three deviations that might be worth considering adding here:

- I defined a Solver type which keeps the solver and its configuration, and the estimators can take an instance or list of instances of it. That let me use the same keyword at the cost of more verbosity, you may not want that, but it may be good to consider expanding what `solver` can take instead of adding new keywords.

- I also added a dictionary to my result type to store all the failures. The successful optimiser and its attributes are stored in the model itself (it's a Julia library based on JuMP) so in my case I didn't save the successful optimiser, but I think doing so here is a good idea.

- In my solution, an in-feasibility certificate doesn't stop the iteration. At least in JuMP, unless the infeasibility is hard, like a weight bound being truly and easily verifiably impossible to hit---eg `lower_bound = 0.75` on all assets when there's only 3 of them and the weights must sum to 1---which short circuits most optimisers, bridges can often relax constraints, transform the problem, or find nearly optimal solutions. The Solver struct has extra bits to let users provide such settings. Also, some optimisers will call a problem infeasible when it's not, so you'd be relying on an optimiser not mis-reporting feasibility. I'd probably have this short-circuit failure as an opt-in for cases like those.

I was checking out the library after posting my feature distance comment. Really nice library.

Issue #301, opened 2026-09-09
Reduce the three F-rated complexity blocks and the two C-band modules


tschm:
[![rhiza](https://img.shields.io/badge/filed_by-rhiza-007ec6)](https://github.com/Jebel-Quant/rhiza-claude)

**Subcategory:** code complexity — score 7 → 9

## Problem

Overall complexity is good: `uvx radon cc src -a` reports **average A (3.39)** across 1887 blocks. The problem is concentrated in the tail — 95 blocks (5.0%) rank C or worse: 75 C, 11 D, 5 E and 4 F.

### Worst blocks by cyclomatic complexity

| Block | CC | Location |
|---|---|---|
| `Attribution.__post_init__` | **F (68)** | `src/skfolio/attribution/_model/_attribution.py:114` |
| `ConvexOptimization._get_weight_constraints` | **F (57)** | `src/skfolio/optimization/convex/_base.py:735` |
| `MeanRisk._fit` | **F (46)** | `src/skfolio/optimization/convex/_mean_risk.py:993` |
| `validate_asset_panel` | **F (44)** | `src/skfolio/utils/validation.py:197` |
| `CharacteristicsFactorModel._fit` | E (38) | `src/skfolio/prior/_characteristics_factor_model.py:991` |
| `rolling_realized_factor_attribution` | E (36) | `src/skfolio/attribution/_realized.py:373` |
| `cross_val_predict` | E (36) | `src/skfolio/model_selection/_validation.py:44` |
| `Population.plot_measures` | E (32) | `src/skfolio/population/_population.py:982` |
| `AssetPanel.info` | E (32) | `src/skfolio/containers/_asset_panel/_panel.py:790` |

### Modules below A on the maintainability index

`uvx radon mi src -s`:

| Module | MI | Lines |
|---|---|---|
| `src/skfolio/prior/_model/_factor_model.py` | **C (2.51)** | 3598 |
| `src/skfolio/attribution/_model/_attribution.py` | **C (6.91)** | 1780 |
| `src/skfolio/containers/_asset_panel/_utils.py` | B (14.32) | 533 |
| `src/skfolio/optimization/convex/_base.py` | B (15.52) | 2561 |
| `src/skfolio/prior/_characteristics_factor_model.py` | B (15.79) | 3210 |

`_factor_model.py` at MI 2.51 across 3598 lines is the single hardest module to change safely.

## Suggested fix

[...]

PR #322, opened 2026-09-16
refactor(attribution): decompose Attribution.__post_init__

tschm:
#### Reference Issues/PRs

Refs #301 — the first of the blocks listed there, done as its own pull request as the
issue asks ("one block per pull request, with the existing test suite as the guard").

`Attribution.__post_init__` ranked **F (68)** on cyclomatic complexity, the worst block
in the package, and carried `_attribution.py` down to a **C (6.91)** maintainability
index — one of the two C-band modules in the issue.

Nearly all of it was a single duplication. A rolling attribution is a stack of
single-point ones: every field is validated identically except that it carries one
extra leading window axis. The two branches therefore repeated the same shape checks,
the same optional-field handling, and the same name-consistency block, with only the
expected shape and some wording differing.

Passing that shape prefix — `(n_windows,)` when rolling, empty otherwise — removes the
split entirely:

```python
leading = (len(self.observations),) if self.is_rolling else ()
```

Validation is now four named helpers plus one method:

| | rank |
|---|---|
| `_check_ndim` / `_check_shape` — the two low-level assertions | A (3) / A (2) |
| `_validate_component` | B (9) |
| `_validate_breakdown` | B (10) |
| `Attribution._validate_asset_by_factor_contrib` | B (8) |
| `Attribution.__post_init__` | **B (6)** |

The field tuples become module constants instead of being rebuilt on every
instantiation.

#### Behaviour is unchanged, including the asymmetries

Every error message is preserved **verbatim**. That includes two asymmetries between
the branches that a tidier refactor would have "fixed" silently, which would be a
behaviour change rather than a refactor:

- only the single-point breakdown message points back at `` `names` ``;
- single-point `mu_uncertainty` is not validated, while the rolling one is.

The `expected` shapes are rendered by `str(tuple)`, which reproduces the hand-written
`f"({n_windows},)"` and `f"({n_windows}, {n_items})"` forms exactly.

`weight_std` keeps its scalar exemption (it is `0.0` for a static allocation) while
`exposure_std` and `mu_contrib_uncertainty` keep theirs, so a scalar there still raises
`AttributeError` exactly as before rather than silently passing.

No new tests: this is a behaviour-preserving refactor, and the suite already pins 19 of
these error messages in `tests/test_attribution/test_model.py`, which is what makes the
verbatim-message requirement checkable. Those tests are untouched and pass.

`350 passed` in `tests/test_attribution`, `7826 passed, 1 skipped` for the full suite,
`ruff check` and `ruff format` clean.

#### Measurements

```
before:  Attribution.__post_init__  F (68)   module MI  C (6.91)
after:   Attribution.__post_init__  B (6)    module MI  B (14.54)
```

The worst block in the file is now the pre-existing `_plot_contribution_chart` at
D (28), which is already within the issue's "no block above D" target.

Remaining for #301 after this: `ConvexOptimization._get_weight_constraints` F (57),
`MeanRisk._fit` F (46), `validate_asset_panel` F (44), four E blocks, and
`prior/_model/_factor_model.py` at MI C (2.51).


HugoDelatte (approved):
Thanks for the PR, it improves maintainability by removing duplicated validation between single-point and rolling attribution.

I've made a few follow-up improvements:
- Replaced abbreviated names with more descriptive ones.
- Removed the generic dimension-check helper and kept those checks beside the relevant domain rules.
- Simplified component validation and optional-field handling.
- Shortened private-helper documentation and clarified comments.

I also fixed a diagnostic regression: single-point asset-by-factor contributions are already two-dimensional, so inferring rolling attribution from the expected rank produced an incorrect “Rolling” error prefix. The regression tests now cover both contribution fields and both attribution modes.

PR #323, opened 2026-09-16
refactor(optimization): decompose the two F-rated convex blocks

tschm:
#### Reference Issues/PRs

Refs #301, continuing after #322 (`Attribution.__post_init__`). Independent of it —
different files, no conflict — but it does touch `_base.py` and `_mean_risk.py`
alongside #321, so whichever lands second needs a small rebase.

`ConvexOptimization._get_weight_constraints` ranked **F (57)** and `MeanRisk._fit`
**F (46)**: the second and third worst blocks in the package. Neither is intricate —
both are long because they are *sequential*, a series of independent stages appending
to one `constraints` list.

**`_get_weight_constraints`** now cleans its inputs and delegates one method per
constraint family:

| method | rank |
|---|---|
| `_bound_constraints` — per-asset bounds, long and short exposure caps | B (9) |
| `_budget_constraints` | B (6) |
| `_mip_constraints` — cardinality and threshold, with the validation that applies only there | C (13) |
| `_linear_constraints` — group and factor equations | B (10) |
| `_inequality_constraints` — the explicit `A @ w <= b` | B (9) |
| `_get_weight_constraints` | **C (15)** |

**`MeanRisk._fit`** hands off four stages:

| method | rank |
|---|---|
| `_set_scales` — per-measure scaling, and the non-1-homogeneous ratio warnings | B (8) |
| `_efficient_frontier_constraint` — return bounds and the parametrized constraint | A (2) |
| `_risk_constraints` — the single pass building the optimized measure and every `max_<measure>` limit | C (14) |
| `_build_objective` — including the Charnes-Cooper transformation | B (10) |
| `MeanRisk._fit` | **C (16)** |

#### What the split had to preserve

The extracted code is moved verbatim — same expressions, same order, same error
messages and warnings. Two couplings were load-bearing and are now explicit rather
than accidental:

- **The MIP solver capability check stays in `_get_weight_constraints`.** It guards the
  whole method, not just the mixed-integer family, and moving it into
  `_mip_constraints` would have changed when it fires.
- **`_risk_constraints` returns its parameters** so the `max_<measure>` limits reach
  `parameters_values` in the original order, which is what the multi-optimization sweep
  indexes by position.

One thing I deliberately did *not* extract: the tracking-error block rebinds `y`
(`y = y[y.columns[0]]`, then `_, y = skv.validate_data(self, X, y)`), and that rebinding
is read later by the efficient-frontier stage. Extracting it would have meant returning
`y` to keep the coupling, which reads worse than leaving the block in place.

No new tests: this is behaviour-preserving, and the existing suite is the guard the
issue asks for. `1993 passed` in `tests/test_optimization`, `7826 passed, 1 skipped`
for the full suite, `ruff check` and `ruff format` clean.

#### A note on the maintainability index

Module MI barely moves here, and slightly **down**: `_base.py` 15.52 → 14.22,
`_mean_risk.py` 34.85 → 33.25, both keeping their rank. That is expected rather than a
regression — MI is driven by module size and *total* complexity, and decomposition
redistributes complexity rather than removing it, while adding signatures and
docstrings. It is worth stating because it means the remaining MI target in #301,
`prior/_model/_factor_model.py` at C (2.51) over 3598 lines, cannot be fixed by
decomposition: only splitting the module or deleting code moves that number.
(`_attribution.py` reached B in #322 because that refactor deleted a large duplication.)

Remaining for #301 after this: `validate_asset_panel` F (44), four E blocks
(`CharacteristicsFactorModel._fit`, `cross_val_predict`,
`rolling_realized_factor_attribution`, `Population.plot_measures`, `AssetPanel.info`),
and the `_factor_model.py` module split.


HugoDelatte (approved):
Thanks the the PR. The new helpers provide a readability and maintainability improvement.

I added a cleanup to clarify four helper names, shorten docstrings to useful contracts, simplify mixed-integer returns, and preserve warning locations. I've also improved typing to follow existing skfolio convention.

PR #324, opened 2026-09-16
refactor(containers): decompose validate_asset_panel and AssetPanel.info

tschm:
#### Reference Issues/PRs

Refs #301, third in the series after #322 (attribution) and #323 (convex). Independent
of both — different files.

Note that `validate_asset_panel` at F (44) is **not** in the issue's table: it has
appeared since the issue was written. `uvx radon cc src -nc` is the acceptance
criterion, so it has to go too.

Two repetitive blocks, not intricate ones.

**`validate_asset_panel` F (44) → C (20).** It repeated the same five-line shape of
check once per value constraint: look the fields up, build a boolean `bad` mask, reduce
a 3-D field over its third axis, restrict to `active_mask` for the two `when_active`
rules, then raise a message naming the first offending observation. Only three things
ever differed — the predicate, the active restriction, and the wording — so they become
a table:

```python
_ValueRule(
    "strictly_positive_when_active",
    lambda values: ~np.isfinite(values) | (values <= 0),
    True,
    "non-finite or non-positive values for active assets",
    "be strictly positive and finite wherever `active_mask` is True.",
)
```

and the body becomes `_check_value_rule`, A (4). The seven
`if x is not None and len(x) == 0: x = None` normalizations collapse to `x or None`,
which is the same thing for `list | None`.

**`AssetPanel.info` E (32) → B (7).** It built one long list of lines across four
sections, one of them via a closure appending to that list. Each section is now a
function returning its own lines: `_info_headline_missing_lines` B (6),
`_info_mask_lines` A (5) (the former closure), `_info_field_coverage_lines` B (8) and
`_info_categorical_lines` C (14).

#### Verified identical, not assumed identical

Neither block is covered by tests that pin its full output, so I checked both against
the pre-refactor implementation directly:

- **Every one of the five rules** was triggered on a panel that violates it, in both the
  2-D and the 3-D field form (8 cases), and the old and new implementations produce
  **byte-identical** `ValueError` messages. This matters because the messages are not
  uniform — `finite_when_active` interpolates the field name three times, and
  `strictly_positive_when_active` wraps mid-sentence.
- **`info()` produces a byte-identical 128-line report** across four panels: all fields
  and both masks set; no masks; estimation mask equal to the active mask (the "Same as
  Active Mask." path); and a 3-D-only panel (the "no 2D field, no headline" path).

No new tests: behaviour-preserving refactor, existing suite as the guard.
`618 passed` in `tests/test_utils` + `tests/test_containers`, `7826 passed, 1 skipped`
for the full suite, `ruff check` and `ruff format` clean.

#### Measurements

```
validate_asset_panel   F (44) -> C (20)      utils/validation.py   MI A (51.85) -> A (59.07)
AssetPanel.info        E (32) -> B (7)       _panel.py             MI A (19.22) -> B (18.61)
```

`utils/validation.py` gains because duplication was deleted. `_panel.py` dips slightly
because its sections were only moved apart — the added signatures and docstrings cost a
little more than the split saves, the same effect noted in #323. Neither crosses out of
the issue's target band.

Remaining for #301 after this: four E blocks (`CharacteristicsFactorModel._fit`,
`cross_val_predict`, `rolling_realized_factor_attribution`, `Population.plot_measures`)
and the `prior/_model/_factor_model.py` module split for its C (2.51) MI.


HugoDelatte (approved):
Thanks for the PR.

I've kept most of the improvements in _panel.py but removed the PR's _ValueRule abstraction which added more complexity than the original, and replaced it with simpler code.

PR #325, opened 2026-09-16
refactor: decompose three E-rated blocks

tschm:
#### Reference Issues/PRs

Refs #301, fourth in the series after #322, #323 and #324. Independent of all three.

Three E-rated blocks, each a pipeline whose stages are already separate in the reader's
head but not in the code.

**`cross_val_predict` E (36) → C (12)**, in four steps:

| | rank |
|---|---|
| `_check_unshuffled_folds` | A (5) |
| `_predict_sequential` — carries holdings fold to fold, parallelizing only across independent paths | B (7) |
| `_predict_independent` | A (1) |
| `_assemble_predictions` — per-fold predictions into a `Population` or one `MultiPeriodPortfolio` | C (12) |

**`rolling_realized_factor_attribution` E (36) → C (13)**, into
`_resolve_uncertainty_inputs`, `_validate_rolling_window`, `_attribution_per_window`
and `_stack_rolling_results`.

**`Population.plot_measures` E (32) → C (20)**. It built three different figures inline
and repeated `{"title": ..., "tickformat": ",.1%" if not measure.is_ratio else None}`
nine times. The figures become `_measure_surface_figure`, `_measure_scatter_3d_figure`
and `_measure_scatter_figure`; the repetition becomes `_measure_axis` and
`_measure_scene`.

#### Verified identical

`plot_measures` produces **byte-identical figure JSON** across seven variants: 2-D, 2-D
with a ratio measure, with a color scale, with non-dominated fronts, 3-D, surface, and
with extra hover measures. The other two are covered by the existing suite.

#### One bug found, deliberately not fixed here

While building that check: **`plot_measures(to_surface=True)` raises `TypeError` when
the z measure is a ratio.** The hovertemplate does

```python
str(e) + ": %{" + v + ":" + (",.3%" if not e.is_ratio else None) + "}"
```

and concatenating `None` to a `str` raises. It fails identically before this PR —
`pop.plot_measures(x=RiskMeasure.VARIANCE, y=PerfMeasure.MEAN, z=RatioMeasure.SHARPE_RATIO, to_surface=True)`
reproduces it on `main`. I left it alone because this PR changes no behaviour; happy to
fix it here or in a separate one, whichever you prefer.

No new tests: behaviour-preserving refactor, existing suite as the guard.
`7826 passed, 1 skipped`, `ruff check` and `ruff format` clean. (The bug above is
untested on `main` too — a regression test would come with the fix.)

#### What is left in #301 after this

Every F block and four of the five E blocks in the issue are now handled across #322,
#323, #324 and this PR. Two items remain, and both need a design decision rather than
mechanical extraction, so I would rather agree the approach with you than guess:

1. **`CharacteristicsFactorModel._fit` E (38).** Unlike the other eight, this one is not
   a sequence of separable stages: it threads roughly thirty interdependent locals
   through one pipeline (exposures, reduced exposures, the lagged variants, the basis,
   the currency counterparts of each, masks, buffers). Every seam I tried needs a helper
   taking or returning six to eight of them, which is worse code bought purely with a
   metric. The honest fix is a fit-context object, or splitting the estimator — a design
   change, and yours to direct.
2. **`prior/_model/_factor_model.py` at MI C (2.51)**, 3598 lines. As #323 noted, MI is
   driven by module size, so decomposition cannot move it: this one needs the module
   split into several files, which changes the package layout and should be your call on
   where the seams go.


HugoDelatte:
I’m closing this PR because I find the original code is easier to read and maintain. Several extractions introduce large parameter lists and additional navigation without removing meaningful duplication.

I think that's another interesting example showing that cyclomatic complexity is a useful diagnostic, but reducing it does not automatically improve maintainability. 

PS: The shared plotting-axis formatting is worthwhile and could be proposed as a smaller, focused change.

PR #326, opened 2026-09-16
refactor(prior): split the factor model module along its method groups

tschm:
#### Reference Issues/PRs

Refs #301, fifth and last in the series after #322, #323, #324 and #325. Independent of
all of them.

This closes the maintainability half of the issue's acceptance criterion:
`uvx radon mi src -s` now reports no module below B.

`prior/_model/_factor_model.py` was **3598 lines at MI C (2.51)** — the worst module in
the package, and the one the issue calls "the single hardest module to change safely".

As #323 and #324 established empirically, MI is driven by module size and *total*
complexity, so decomposing a block cannot move it: splitting the module is the only
lever. You asked where the seams should go — I used the ones the class already
documents. `FactorModel`'s docstring describes four method groups, and the file marked
each with a section comment (`# Idiosyncratic diagnostics`,
`# Cross-sectional regression diagnostics`, `# Exposure diagnostics`, `# Attribution`),
so each group becomes a module and a mixin:

| module | mixin | MI |
|---|---|---|
| `_factor_model_idio.py` | `_IdioDiagnosticsMixin` | A (65.38) |
| `_factor_model_regression.py` | `_CrossSectionalRegressionDiagnosticsMixin` | A (64.13) |
| `_factor_model_exposure.py` | `_ExposureDiagnosticsMixin` | A (54.85) |
| `_factor_model_attribution.py` | `_AttributionMixin` | A (86.33) |
| `_factor_model_plots.py` — plotting primitives shared by several groups | — | A (58.24) |
| `_factor_model_utils.py` — statistics, selector and NamedTuple helpers | — | A (59.52) |
| `_factor_model.py` — fields, factor-structure methods, private helpers | — | **A (25.99)** |

#### Why mixins

Rather than free functions with delegating methods: mixins keep each method's docstring
next to its code and need no boilerplate. A plain mixin is also compatible with the
frozen dataclass — `@dataclass(frozen=True)` only rejects *non-frozen dataclass* bases,
and the mixins carry no fields, so they do not touch the generated `__init__` or the
MRO's field order. `cached_property` keeps working for the reason it already did: it
writes through `__dict__`, not `__setattr__`.

#### Nothing else changes

- `skfolio.prior.FactorModel` and `skfolio.prior.CorrelationMethod` keep their import
  paths, and `prior/_model/__init__.py` is untouched.
- `FactorModel.__module__` is still `skfolio.prior._model._factor_model`, so existing
  pickles stay valid — verified with a round-trip.
- The only import that had to move is in a test reaching for two private helpers
  (`_exceedance_agg`, `_selector_to_positions`), now in `_factor_model_utils`.

#### Verified a pure move

Rather than eyeballing a 2477/2270 diff, I compared the **AST of every definition**
before and after — all 83 methods, module functions and NamedTuple classes:

```
old definitions: 83   new definitions: 83
missing: []   added: []   changed: []
PURE MOVE
```

So the only edits are the import blocks, the four class statements and one docstring
sentence pointing at the new layout.

No new tests: nothing but the file layout changed. `925 passed` in `tests/test_prior`,
`7826 passed, 1 skipped` for the full suite, `ruff check` and `ruff format` clean.

#### Where #301 stands after this

- `uvx radon mi src -s`: **no module below B.** ✅
- `uvx radon cc src -nc`: no block above D **except `CharacteristicsFactorModel._fit`
  E (38)**, once #322–#325 land.

That last block is the one I flagged in #325: it is not a sequence of separable stages
but roughly thirty interdependent locals threaded through one pipeline, and every seam
costs a helper taking six to eight of them. It wants a fit-context object or an
estimator split — a design change I would rather agree with you first.


HugoDelatte:
I’m closing this PR because I find the original class structure easier to understand and maintain.

The extracted mixins still depend extensively on `FactorModel`’s fields, cached properties, and private helpers. The same closely connected implementation is now spread across multiple files, with four inheritance bases added. Understanding a diagnostic still requires following its dependencies back into the main class.

The method groups are reasonable, but smaller files and higher Maintainability Index scores do not, by themselves, establish a clearer design. I would prefer to keep the stateful methods and their shared calculations together, while considering a smaller, focused change for self-contained plotting helpers..

PR #327, opened 2026-09-16
refactor(prior): thread CharacteristicsFactorModel._fit through a fit state

tschm:
#### Reference Issues/PRs

Refs #301, and the design change I flagged in #325. Sixth and last in the series after
#322, #323, #324, #325 and #326. Independent of all of them.

With this, `uvx radon cc src -nc` reports **no block above D** across the package, which
together with #326 closes both halves of the issue's acceptance criterion.

`CharacteristicsFactorModel._fit` ranked **E (38)** and was the one block in #301 that
did not decompose into helpers. Unlike the other eight, it is not a sequence of
separable steps: it is a pipeline threading roughly thirty interdependent locals
(exposures, reduced exposures, the lagged variants of each, the basis, their currency
counterparts, masks, buffers), and every seam I tried wanted a helper taking six to
eight arrays — worse code bought with a metric.

So the arrays become the thing they already were: state. `_FitState` is a private
dataclass carrying what one stage leaves for the next, and `_fit` becomes the pipeline
it always was:

```python
state = self._load_panel(...)
self._trim_warmup(state)
self._reduce_exposures(state)
self._lag_exposures(state, first_call=first_call, method=method)
self._regress(state, routed_params=routed_params)
self._add_currency_factors(state)
self._estimate_moments(...)
self._reduce_to_investment_universe(state)
self._assemble_return_distribution(state, first_call=first_call)
```

The stage names are the ones the section comments already used, so this is the
structure the file described rather than a new one. Each stage takes `(self, state)`
plus only what it genuinely needs beyond the arrays — the panel, the routed params,
`first_call`, `method`. `_FitState`'s fields are grouped and commented by the stage
that fills them, so the data flow is readable instead of inferred across four hundred
lines.

| | rank |
|---|---|
| `_load_panel` | B (6) |
| `_trim_warmup` | A (5) |
| `_reduce_exposures` | A (4) |
| `_lag_exposures` | B (10) |
| `_regress` | A (1) |
| `_add_currency_factors` | A (5) |
| `_estimate_moments` | A (1) |
| `_reduce_to_investment_universe` | A (4) |
| `_assemble_return_distribution` | B (9) |
| **`_fit`** | **E (38) → A (2)** |

#### How the move was made safe

The statements move verbatim — only the threaded locals are renamed to `state.<name>`.
That rename was applied **through the AST**, not with a regular expression, so that
keyword argument names were not caught by it: a textual `\bexposures\b` substitution
would have turned `_validate(exposures=exposures)` into `_validate(state.exposures=…)`,
and there are dozens of such calls.

#### Verified bit-identical, not assumed

The real guard is numerical. Thirteen fitted scenarios, chosen to cover every branch
the stages contain, produce **bit-identical digests for every array** of `factor_model_`
and `return_distribution_` (23 arrays each, 286 digest lines, zero differences):

- `fit`: default, `exposure_lag=3`, constrained families, neutralization, unweighted
  regression (`regression_mcap_power=0`), thresholded idiosyncratic correlation,
  shrunk alpha, bounded `max_history`
- `partial_fit` in three batches: default, constrained families, `exposure_lag=3` —
  these exercise the cross-batch buffers (`_buffer_exposures_reduced`,
  `_buffer_constraint_ratios`, `_buffer_market_cap`), which is where a
  reordering mistake would show up first
- currency factors, in both `fit` and `partial_fit`

No new tests: the behaviour is unchanged and the equivalence check above is a
throwaway harness rather than something worth pinning in the suite. `7826 passed,
1 skipped`, `ruff check` and `ruff format` clean.

If you would like the harness kept as a regression test, say so and I will adapt it —
it would need fixed expected digests, which are platform-sensitive for floating point,
so I left it out rather than add a fragile test.

#### Measurements

```
CharacteristicsFactorModel._fit   E (38) -> A (2)
module MI                         B (15.79) -> B (12.87)
```

Module MI dips for the reason established in #323 and #324: decomposition
redistributes complexity while adding signatures and docstrings. It keeps its rank, and
no module in the package is below B after #326.

#### Where #301 stands

Both acceptance criteria are met once this series lands:

- `uvx radon cc src -nc`: no block above D ✅
- `uvx radon mi src -s`: no module below B ✅


HugoDelatte:
I think that this is a good example where sometimes improving cyclomatic complexity also improves readability and maintainability like in https://github.com/skfolio/skfolio/pull/322, and sometimes it does not like in this PR.

As a human dev, I actually find the original implementation easier to read and maintain. I find that the large mutable fit-state object hides dependencies and makes the data flow harder to follow.

But because I may be biased, I've also asked ChatGPT Astra 6 Ultra which version he prefers and here is the answer:

"""
I would prefer the original implementation. The refactor makes `_fit` easier to skim, but replaces explicit local data flow with a 35-field mutable state object. Understanding each stage now requires tracing which fields previous stages populated, modified, or left obsolete. Some fields also retain earlier array shapes while others are transformed, so the shared object does not provide a clear consistency guarantee.

The shorter methods therefore hide dependencies without simplifying them. Cyclomatic complexity is useful for identifying code that deserves closer review, but it is an imperfect proxy for readability and maintainability. Lowering the score is worthwhile only when the resulting code becomes easier to understand and change.
"""

WDYT? Is there a way to loosen your tool so it does not try to over optimize on cyclomatic complexity alone?

PR #330, opened 2026-09-17
fix(population): render a ratio measure in the surface hover template

### LINKED ISSUE #329: [BUG] plot_measures(to_surface=True) raises TypeError when the z measure is a ratio
[rhiza](https://github.com/Jebel-Quant/rhiza-claude)

**Describe the bug**

`Population.plot_measures(to_surface=True)` raises `TypeError` whenever the `z` measure
is a ratio.

The surface branch builds its hovertemplate by string concatenation, and reuses the
`is_ratio` idiom that the rest of the method uses for plotly's `tickformat` key:

```python
str(e) + ": %{" + v + ":" + (",.3%" if not e.is_ratio else None) + "}"
```

`None` is a valid *value* for `tickformat`, where it means "unset", but it cannot be
concatenated to a string. Every non-ratio measure works; any ratio raises.

Only the surface path is affected. The 2D and 3D scatter paths put the format in a dict
rather than a string, so `None` is harmless there.

**To Reproduce**

```python
import numpy as np
from skfolio import Population, Portfolio, PerfMeasure, RatioMeasure, RiskMeasure
from skfolio.datasets import load_sp500_dataset

X = prices_to_returns(load_sp500_dataset()).iloc[-300:]
rng = np.random.default_rng(0)
population = Population(
    [
        Portfolio(X=X, weights=w / w.sum(), name=f"p{i}")
        for i, w in enumerate(rng.random((5, X.shape[1])))
    ]
)
population.plot_measures(
    x=RiskMeasure.VARIANCE,
    y=PerfMeasure.MEAN,
    z=RatioMeasure.SHARPE_RATIO,   # any ratio measure
    to_surface=True,
)
```

```
File "src/skfolio/population/_population.py", line 1096, in plot_measures
    + (",.3%" if not e.is_ratio else None)
TypeError: can only concatenate str (not "NoneType") to str
```

Swapping `z` for a non-ratio measure such as `RiskMeasure.CVAR` works, as does
`to_surface=False` with the same ratio.

**Expected behavior**

The surface renders, with the ratio hover value shown unformatted rather than as a
percentage — the same distinction the axes already make. Concretely, the ratio branch
needs a format string instead of `None`, e.g. `",.3f"`:

```python
+ (",.3%" if not e.is_ratio else ",.3f")
```

**Additional context**

Found while building an output-equality check for #325, which refactors
`plot_measures` into per-figure helpers. That PR preserves the behaviour verbatim,
including this bug, so it is reported separately rather than fixed there. The line
moves to `_measure_surface_figure` if #325 lands first; the fix is the same either way.

No test covers the surface path with a ratio measure, so a regression test should come
with the fix.

**Versions**

<details>

Reproduced on `main` (`bcca258`, v1.2.8).

</details>

tschm:
#### Reference Issues/PRs

Fixes #329.

Branched from `main` rather than stacked on #325, so the fix is not gated behind a
refactor. It does touch the same lines #325 moves into `_measure_surface_figure`, so
whichever merges second needs a rebase — I will handle it either way; the fix is one
expression and moves unchanged.

`plot_measures(to_surface=True)` raised `TypeError: can only concatenate str (not
"NoneType") to str` whenever the `z` measure was a ratio.

The surface hovertemplate is built by string concatenation, and reused the `is_ratio`
idiom the rest of the method applies to plotly's `tickformat` key:

```python
+ (",.3%" if not e.is_ratio else None)
```

`None` is a valid *value* for `tickformat`, where it means "leave the format unset",
but it cannot be concatenated into a template string. The surface is the only place
that format reaches a string rather than a dict — the nine other `is_ratio` uses in the
method are all dict values — which is why the 2D and 3D scatter paths are unaffected.

#### The fix

```python
+ ("" if e.is_ratio else ":,.3%")
```

A ratio is dimensionless, so this keeps the intent `None` was expressing and drops the
format specifier entirely, leaving plotly to use its default — exactly what the axes and
the colorbar already do for a ratio, and what `plot_contribution` and
`plot_rolling_measure` do by only setting `tickformat` when `not measure.is_ratio`.

I proposed `",.3f"` in the issue; an omitted specifier is more faithful, because it
renders a ratio the same way in the hover, on the axis ticks and on the colorbar,
instead of introducing a third convention. Happy to switch if you prefer an explicit
format.

Resulting templates:

```
ratio z      Variance: %{x:,.3%}<br>Mean: %{y:,.3%}<br>Sharpe Ratio: %{z}<extra></extra>
non-ratio z  Variance: %{x:,.3%}<br>Mean: %{y:,.3%}<br>CVaR: %{z:,.3%}<extra></extra>
```

**Non-ratio measures are unchanged byte for byte** — `":" + ",.3%"` and `":,.3%"` are
the same string, so only the previously-crashing path behaves differently.

One, `test_population_plot_measures_with_a_ratio_measure`, parametrized over
`to_surface`. It asserts the ratio reaches the hover without a format specifier, that a
non-ratio axis keeps `:,.3%`, and that the colorbar format stays unset.

I checked it is not vacuous: with the source change stashed, it fails with the original
`TypeError` at `_population.py:1096`.

The existing `plot_measures` tests only ever passed non-ratio measures, which is how
this survived — the parametrized `to_surface` case right above the new test uses
`MAX_DRAWDOWN` for `z`.

Full suite: `7828 passed, 1 skipped`, `ruff check` and `ruff format` clean.


HugoDelatte (approved):
Thanks for the fix!

PR #338, opened 2026-09-22
fix(portfolio): reject MultiPeriodPortfolio mutations that misalign sample_weight

### LINKED ISSUE #334: [BUG] MultiPeriodPortfolio mutations leave sample_weight misaligned with observations
**Describe the bug**

A valid weighted `MultiPeriodPortfolio` can become invalid after a successful public container mutation. `append()`, deletion, replacement with a different-length period, and assignment to `portfolios` change the observation count without updating or validating `sample_weight`.

The mutation succeeds, but subsequently evaluating `mean` returns `NaN` with a warning about incompatible matrix dimensions.

**To Reproduce**

```python
import warnings
import pandas as pd
from skfolio import MultiPeriodPortfolio, Portfolio

first = Portfolio(
    pd.DataFrame({"asset": [0.01, 0.02]},
                 index=pd.date_range("2026-01-01", periods=2)),
    weights=[1.0],
)
second = Portfolio(
    pd.DataFrame({"asset": [0.03, 0.04]},
                 index=pd.date_range("2026-01-03", periods=2)),
    weights=[1.0],
)
portfolio = MultiPeriodPortfolio([first], sample_weight=[0.25, 0.75])
print("mean before:", round(float(portfolio.mean), 6))

portfolio.append(second)
print("observations:", len(portfolio.returns))
print("sample weights:", portfolio.sample_weight.tolist())
with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter("always")
    print("mean after:", portfolio.mean)
print("warnings:", len(caught))
```

Observed output:

```text
mean before: 0.0175
observations: 4
sample weights: [0.25, 0.75]
mean after: nan
warnings: 1
```

The warning reports failure to calculate `mean` because matrix dimensions differ (four returns versus two weights).

**Expected behavior**

A successful mutation should preserve the one-dimensional, observation-aligned, normalized sample-weight contract. If a mutation cannot infer new weights, it should reject the operation before changing the object, or provide an explicit way to update observations and weights together. It should not guess new observation weights silently.

**Additional context**

This is distinct from merged PR [#315](https://github.com/skfolio/skfolio/pull/315), which validates constructor weights. The example above starts with valid constructor weights; the mismatch is introduced afterwards.

Local public-API regressions reproduce the same invariant violation for:

- `append(new_period)`;
- `del portfolio[index]`;
- `portfolio[index] = shorter_period`;
- `portfolio.portfolios = shorter_list`.

Both `_set_portfolios()` and `append()` change `returns` and `observations` without reconciling the existing weights. The `sample_weight` setter itself correctly rejects a wrong-length assignment.

**Versions**

- skfolio 1.2.9, main commit `c99fcf71349e2df4a7a1033ee85ca2e9ced9abee`
- Python 3.12.13; macOS 26.6.2, arm64
- NumPy 2.5.3; pandas 3.0.6; SciPy 1.18.1; scikit-learn 1.9.1

Vinayak19112003:
Closes #334.

A weighted `MultiPeriodPortfolio` could be silently corrupted by the public mutation API: `append()`, `del`, item replacement, and the `portfolios` setter all changed the observation count without reconciling `sample_weight`. The mutation succeeded, but afterwards measures like `mean` returned `NaN` with a dimension-mismatch warning (exact repro in the issue).

This change makes `_set_portfolios()` and `append()` raise `ValueError` before changing any state when the new observation count does not match the length of the current `sample_weight`, so a weighted portfolio can never end up misaligned. Mutations on unweighted portfolios behave exactly as before, and replacing a period with a same-length one still works. To mutate a weighted portfolio, update `sample_weight` first (e.g. set it to `None`) and assign the new weights afterwards — the error message spells this out.

Regression tests cover all four mutation paths plus the still-valid cases (unweighted append, same-length replacement). Verified: full `test_multi_period_portfolio.py` suite (317 tests) passes, the walk-forward / previous-weights propagation tests in `test_model_selection/test_validation.py` that exercise the internal `portfolios` setter path pass, and ruff lint + format are clean.


HugoDelatte (dismissed):
Thanks for the PR!

I extended the fix to preserve sample-weight inheritance.

The resulting behavior is:

- `sample_weight=None` inherits child weights. Each period’s total weight is proportional to its observation count, and unweighted children contribute equal weights per observation.
- Explicit parent weights override inheritance. Supported mutations preserve them when lengths match and reject incompatible updates before changing the parent.
- Weighted measures exclude NaN returns and renormalize the remaining weights per column.

HugoDelatte (approved):

PR #339, opened 2026-09-22
fix(portfolio): reject shared boundary observation in MultiPeriodPortfolio.append

### LINKED ISSUE #333: [BUG] MultiPeriodPortfolio.append accepts a shared boundary observation that the constructor rejects
**Describe the bug**

With `check_observations_order=True`, constructing a `MultiPeriodPortfolio` with two periods sharing a boundary observation raises `ValueError`. Constructing it with the first period and then appending the second accepts the same input. The resulting return series counts that observation twice.

**To Reproduce**

```python
import pandas as pd
from skfolio import MultiPeriodPortfolio, Portfolio

first = Portfolio(
    pd.DataFrame({"asset": [0.01, 0.02]},
                 index=pd.date_range("2026-01-01", periods=2)),
    weights=[1.0],
)
second = Portfolio(
    pd.DataFrame({"asset": [0.02, 0.03]},
                 index=pd.date_range("2026-01-02", periods=2)),
    weights=[1.0],
)

try:
    MultiPeriodPortfolio([first, second], check_observations_order=True)
except ValueError:
    print("constructor: overlap rejected")

portfolio = MultiPeriodPortfolio([first], check_observations_order=True)
portfolio.append(second)
print("append: accepted")
print(pd.DatetimeIndex(portfolio.observations).strftime("%Y-%m-%d").tolist())
print(portfolio.returns.tolist())
```

Observed output:

```text
constructor: overlap rejected
append: accepted
['2026-01-01', '2026-01-02', '2026-01-02', '2026-01-03']
[0.01, 0.02, 0.02, 0.03]
```

**Expected behavior**

`append()` should reject the shared boundary just as the constructor does, before changing the portfolio. A period starting strictly after the preceding period should remain accepted.

**Additional context**

In `_multi_period_portfolio.py`, `_set_portfolios()` compares the boundary with `<=`, whereas `append()` uses `<`.

The existing append-order test covers reversed and disjoint periods, but not equality at the boundary. Local public-API checks confirm constructor rejection and disjoint append acceptance; the equivalent append-rejection regression fails.

**Versions**

- skfolio 1.2.9, main commit `c99fcf71349e2df4a7a1033ee85ca2e9ced9abee`
- Python 3.12.13; macOS 26.6.2, arm64
- NumPy 2.5.3; pandas 3.0.6; SciPy 1.18.1; scikit-learn 1.9.1

Vinayak19112003:
Closes #333.

With `check_observations_order=True`, the constructor rejects two periods sharing a boundary observation, but `append()` accepted the same input — the shared observation ended up counted twice in the return series (repro in the issue).

Root cause is the one the issue names: `_set_portfolios()` compares the boundary with `<=` while `append()` used `<`. This aligns `append()` on `<=` so both paths reject the same overlapping input with the same error, before any state is mutated. A period starting strictly after the previous period is still accepted.

The new regression test covers constructor rejection, append rejection (and that the failed append leaves the object untouched), and acceptance of a strictly-after period. The append/order tests and ruff lint + format pass.

Note: this touches the same `append()` method as my PR #338 (for #334); the two changes are independent and conflict-free in content (one adds a `sample_weight` guard, the other tightens the boundary comparison).


HugoDelatte:
Fixed in https://github.com/skfolio/skfolio/pull/338

